PR #109, opened 2022-08-27
Fix/excessive leakage

khemrajrathore:
fix - remove leakages which don't contain any variable


pandurangpatil (approved):
looks good to me.

PR #110, opened 2022-08-29
request: sync-retry changes for cli

abhstabs:


src/main/scala/ai/privado/entrypoint/CommandParser.scala
```diff
@@ -111,6 +113,19 @@ object CommandParser {
                 failure("external rule files location is required if you ignore the internal rules")
               else success
             )
+          ),
+        cmd(CommandConstants.RETRY_SYNC)
+          .required()
+          .action((_, c) => c.copy(cmd = c.cmd + CommandConstants.RETRY_SYNC))
```
khemrajrathore, line 119: Add a text, which describes the command

src/main/scala/ai/privado/entrypoint/SyncProcessor.scala
```diff
@@ -0,0 +1,30 @@
...
+import better.files.File
+import org.slf4j.LoggerFactory
+
+object SyncProcessor extends CommandProcessor {
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  def checkOutputFileExists(): Boolean = {
+    val repoPath       = config.sourceLocation.head
+    val outputFileName = "privado"
+    File(repoPath + "/.privado/" + outputFileName + ".json").exists
```
khemrajrathore, line 15: @abhstabs I see we have used the output folderName and file name `.privado` and `privado` across, Let's create a  constant for them, so they are not repeated

src/main/scala/ai/privado/entrypoint/CommandParser.scala
```diff
@@ -51,10 +51,12 @@ object CommandConstants {
   val DISABLE_DEDUPLICATION_ABBR = "dd"
   val IGNORE_EXCLUDE_RULES       = "ignore-exclude-rules"
   val IGNORE_EXCLUDE_RULES_ABBR  = "ier"
+  val RETRY_SYNC                 = "retry-sync"
```
khemrajrathore, line 54: @abhstabs The command name `retry-sync` doesn't relate directly to uploading the `privado.json` and metrics. 
we can give a thought on `privado upload` or `privado sync-result`, `privado upload-result`
cc - @pandurangpatil

src/main/scala/ai/privado/model/Constants.scala
```diff
@@ -94,15 +94,17 @@ object Constants {
...
+  val cliVersion          = "privadoCLIVersion"
+  val mainVersion         = "privadoMainVersion"
+  val notDetected         = "Not Detected"
+  val createdAt           = "createdAt"
+  val repoName            = "repoName"
+  val gitMetaData         = "gitMetaData"
+  val sources             = "sources"
+  val localScanPath       = "localScanPath"
+  val processing          = "processing"
+  val outputFileName      = "privado"
```
khemrajrathore, line 107: @abhstabs we can have outputFileName as `privado.json`. What is your thought?

pandurangpatil:
@abhstabs It seems you have not gone through the ticket properly. I was visualizing it to be a common option to retry failed uploads. However, if this is what a common perception of retrial will be then we should continue with your implementation. 

I agree with @khemrajrathore comment with regards to the command name `privado upload-result` looks good.

abhstabs:
@pandurangpatil I had a discussion with @khemrajrathore about this yesterday. Having `upload-result` as a command can be perceived as having to run `scan` to scan the repository and then using `upload-result` in case we want to view the results on the UI dashboard. `retry-sync` or `sync-result` is better worded as that removes the ambiguity. You can let us know which should we proceed with. 

Also for the upload part, since the results file would be getting overwritten by every scan, only uploading the latest file with the newest results should be sufficient. Do we need to change anything there?

pandurangpatil:
@abhstabs I guess `retry-upload-result` will be more relevant as having `retry` in the command will make sure to pass the message that this command is for retrial. 

Uploading the latest file will be sufficient.

khemrajrathore (approved):
looks good to me

PR #120, opened 2022-09-08
semver: replaced . with - for versioning in dev

hiteshbedre:


pandurangpatil (approved):
As docker tags don't allow it to have "+" sign, approving changes for "-" for now.

PR #124, opened 2022-09-10
fix: stringify json before request, change key names in metrics

abhstabs:
Hi @khemrajrathore / @ojaswa1942, 

Changes to Stringify the Json before uploading to cloud. Also changed the key names as request by Vineet. Please have a look. 

Thank you.

cc: @pandurangpatil


ojaswa1942:
Since it looks like we've added an external dependency, and we're still using the `ScalaDefautModule` - did we simply try `toString` to stringify?

abhstabs:
`toString` only converts the Json object to a string. Doesn't actually _stringify_ the data that we were looking for. @khemrajrathore and I tried that yesterday.

khemrajrathore (approved):
LGTM

ojaswa1942 (approved):
:shipit:

PR #126, opened 2022-09-12
change: update metrics for scan and upload status

abhstabs:
Hi @ojaswa1942 / @khemrajrathore , 

added metrics for upload and scan status. Please have a look. 

Thank you. 

cc: @pandurangpatil

### DIFF
--- a/src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
+++ b/src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
@@ -313,9 +313,12 @@ object ScanProcessor extends CommandProcessor {
         val dataflowMap = cpg.dataflow
 
         println("Brewing result...")
+        MetricHandler.setScanStatus(true)
         // Exporting
         JSONExporter.fileExport(cpg, outputFileName, sourceRepoLocation, dataflowMap) match {
-          case Left(err) => Left(err)
+          case Left(err) =>
+            MetricHandler.otherErrorsOrWarnings.addOne(err)
+            Left(err)
           case Right(_) =>
             println(s"Successfully exported output to '${AppCache.localScanPath}/$outputDirectoryName' folder")
             logger.debug(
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -328,6 +331,7 @@ object ScanProcessor extends CommandProcessor {
       case Failure(exception) => {
         logger.error("Error while parsing the source code!")
         logger.debug("Error : ", exception)
+        MetricHandler.setScanStatus(false)
         Left("Error while parsing the source code: " + exception.toString)
```
ojaswa1942, line 335: Do we capture this error message outside of this function?
If not, we should capture the error too

abhstabs: We capture the error outside of this function so it's not included here.

ojaswa1942: @abhstabs Should not both metrics, for one thing, be handled together.. if that makes the code cleaner?

abhstabs: The errors are logged one by one as they occur on a lower level. This is an umbrella exception to handle the user side messages.

ojaswa1942: @abhstabs Makes sense. 
Just want to ensure that even "unhandled exceptions" that are raised from lower-level functions are caught, which might not be handled anywhere else but only in the top handler. Looks good if you can confirm that.

abhstabs: Can confirm. Scan errors and upload errors are reported in the metrics at the lower level.

ojaswa1942 (approved):
Since all unhandled exceptions are also reported, as confirmed by @abhstabs, looks good.

pandurangpatil:
@khemrajrathore have look at this one and if looks good to you go ahead and merge. @ojaswa1942 has already approved this one.

khemrajrathore (approved):
LGTM

PR #137, opened 2022-09-20
structure refactoring

khemrajrathore:
Refactoring the RuleInfo remains, as new ruleStructures will be needed for JS


pandurangpatil:
@khemrajrathore have you tested the one or two repos with this code refactor?

src/main/scala/ai/privado/semantic/Language.scala
```diff
@@ -22,18 +22,10 @@
 
 package ai.privado.semantic
```
pandurangpatil, line 24: @khemrajrathore did we create a new Language.scala as a subpackage inside java package?

khemrajrathore: yes, we had this earlier outside, now we have segregated into 2, generic ones outside, lang specific inside java

pandurangpatil:
@ojaswa1942 If you find the changes good, go ahead and merge the changes.

ojaswa1942 (approved):

PR #138, opened 2022-09-20
add - processing language specific rule

khemrajrathore:
- Process only language specific rules. (Make use of the `language.yaml`)
Ex - If scanning for `javascript`, `java` rules should not be considered
- Cache relevant rules only


pandurangpatil (approved):
looks good to me.

PR #139, opened 2022-09-21
add - JsSrc2Cpg

khemrajrathore:
add - JsSrc2Cpg
patch - filtering rules by language after detecting language
add - sample tagger integration


pandurangpatil (approved):
Looks good to me.

PR #144, opened 2022-09-22
Add/refactoring by lang

khemrajrathore:


src/main/scala/ai/privado/javascript/passes/methodfullname/MethodFullNameFromIdentifier.scala
```diff
@@ -0,0 +1,95 @@
...
+    * log4js.configure({ appenders: { cheese: { type: "file", filename: "cheese.log" } }, categories: { default: {
+    * appenders: ["cheese"], level: "error" } }, });
+    *
+    * const logger = log4js.getLogger("cheese");
+    *
+    * logger.debug("Information is available);
+    *
+    * We are looking to add methodFullName to `debug`
+    *
+    * In this case the methodFullName will be pkg.log4js.getLogger.debug
```
pandurangpatil, line 52: @khemrajrathore I guess then we can also take care of the following scenario that we were discussing in the afternoon.

```
    const users = require("../controllers/accounts.controller.js");
  
    var router = require("express").Router();
```

khemrajrathore: ya we can try, the query for this will be slight different

src/main/scala/ai/privado/javascript/passes/methodfullname/MethodFullNameFromIdentifier.scala
```diff
@@ -0,0 +1,99 @@
...
```
pandurangpatil, line 52: @khemrajrathore I guess we need to update the doc comments here?

pandurangpatil (approved):
looks good to me.

PR #148, opened 2022-09-26
Support for curly brace imports

khemrajrathore:
add - support for  `const {WebClient}  = require('@slack/web-api')`
add - support for adding methodFullName in chaining via fieldAccess - Ex- `web.chat.postMessage()`


pandurangpatil (approved):
looks good to me.

PR #152, opened 2022-09-28
patch: add exception handling for guessLanguage

abhstabs:
Fix for handling the exception in the issue https://github.com/Privado-Inc/privado/issues/76 

Please have a look. Thank you. 

cc: @ojaswa1942 / @pandurangpatil

### DIFF
--- a/src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
+++ b/src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
@@ -260,33 +260,34 @@ object ScanProcessor extends CommandProcessor {
     AppCache.init(sourceRepoLocation)
     import io.joern.console.cpgcreation.guessLanguage
     println("Guessing source code language...")
-    var languageDetected: Option[String] = None
-    try {
-      languageDetected = guessLanguage(sourceRepoLocation)
-    } catch {
-      case exc: Exception =>
-        println(s"Error Occurred: ${exc.toString}")
-        exit(1)
-    }
-    val xtocpg = languageDetected match {
-      case Some(lang) =>
-        if (!(lang == Languages.JAVASRC || lang == Languages.JAVA)) {
-          if (checkJavaSourceCodePresent(sourceRepoLocation)) {
-            println(s"We detected presence of 'Java' code base along with other major language code base '${lang}'.")
-            println(s"However we only support 'Java' code base scanning as of now.")
-          } else {
-            println(s"As of now we only support privacy code scanning for 'Java' code base.")
-            println(s"We detected this code base of '${lang}'.")
-            exit(1)
+    val xtocpg = Try(guessLanguage(sourceRepoLocation)) match {
+      case Success(languageDetected) =>
+        languageDetected match {
+          case Some(lang) =>
+            if (!(lang == Languages.JAVASRC || lang == Languages.JAVA)) {
+              if (checkJavaSourceCodePresent(sourceRepoLocation)) {
+                println(
+                  s"We detected presence of 'Java' code base along with other major language code base '${lang}'."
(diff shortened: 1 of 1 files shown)


src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -260,7 +260,15 @@ object ScanProcessor extends CommandProcessor {
...
+    var languageDetected: Option[String] = None
+    try {
+      languageDetected = guessLanguage(sourceRepoLocation)
+    } catch {
+      case exc: Exception =>
+        println(s"Error Occurred: ${exc.toString}")
+        exit(1)
+    }
+    val xtocpg = languageDetected match {
       case Some(lang) =>
```
khemrajrathore, line 272: @abhstabs A better way to write this can be, what is your thought?

```
val xtocpg  = Try(guessLanguage(sourceRepoLocation)) match {
 case Success(langDetected) => 
     langDetected match {
         case Some(lang) => ....
         case _ => .....
     }
    case Failure(exc) => 
    println(s"Error Occurred: ${exc.toString}")
       exit(1)

}
```

src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -260,7 +260,15 @@ object ScanProcessor extends CommandProcessor {
     AppCache.init(sourceRepoLocation)
     import io.joern.console.cpgcreation.guessLanguage
     println("Guessing source code language...")
-    val xtocpg = guessLanguage(sourceRepoLocation) match {
+    var languageDetected: Option[String] = None
+    try {
+      languageDetected = guessLanguage(sourceRepoLocation)
+    } catch {
+      case exc: Exception =>
+        println(s"Error Occurred: ${exc.toString}")
```
pandurangpatil, line 268: @abhstabs You are printing the error stack trace. I guess you need to show an error message (by calling getMessage() method on exception) instead of the stack trace. And log the entire stack trace in debug log the way we are doing it everywhere else. (We need to make sure the moment we catch any exception, the first job is to log the error in debug mode.)

pandurangpatil (changes requested):
Make the respective change.

khemrajrathore (approved):
LGTM

ojaswa1942:
Partially resolves https://github.com/Privado-Inc/privado/issues/76

src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -260,25 +260,34 @@ object ScanProcessor extends CommandProcessor {
...
-        } else {
-          println(s"Detected language 'Java' ")
         }
-        createJavaCpg(sourceRepoLocation, lang)
-      case _ => {
-        logger.error("Unable to detect language! Is it supported yet?")
-        Failure(new RuntimeException("Unable to detect language!"))
-      }
+      case Failure(exc) =>
+        logger.debug(exc.toString)
```
pandurangpatil, line 288: @abhstabs small correction you need to make logger.debug("Error while guessing language", exc)

pandurangpatil: exc.toString will not print the stack trace.

pandurangpatil (changes requested):
Make this small change.

src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
abhstabs: Updated. Please have a look.

pandurangpatil (approved):
Looks good to me

PR #156, opened 2022-09-30
add - test case for methodFullName pass

khemrajrathore:
note - testCase failing as package.json is not read currently in test, Have raised a query for that, once that is responded back will make respective changes


pandurangpatil:
@khemrajrathore I don't see any tests

pandurangpatil:
@khemrajrathore 

The following use cases need to be added.

1. `const moment require 'moment';`
2. `const internalConst require '../../interanal';` - If we are not handling the internal imports. We should add an empty test case along with details of why we haven't handled it, along with TODO: item.
3. use cases with `import` keyword for similar combinations could more as well refer - https://nodejs.org/api/esm.html#node-imports.

pandurangpatil (approved):
looks good to me.

PR #164, opened 2022-10-07
patch - change createCpg to createCpgWithAllOverlays

khemrajrathore:
add - dependency upgrade to "1.1.1134"
patch - change `cpg.staticImports` to `cpg.imports`


pandurangpatil (approved):
looks good to me.

PR #165, opened 2022-10-07
fix - proper filename for getting dump in excerpt

khemrajrathore:
fix - blank excerpt in docker image
fix - wrong repoName in privado.json for windows machine
patch - error message to inform user if something goes wrong in scan


pandurangpatil (approved):
looks good to me.

PR #174, opened 2022-10-12
Feature/rule validator

khemrajrathore:
fix - validation of rule failing on exclusion and config (because config directory was getting skipped)
add - adding ruleValidation for config dir


pandurangpatil (approved):
looks good to me.

PR #183, opened 2022-10-28
add - collection point for JS

khemrajrathore:
Covers scenario where the handler code is present in the endpoint declaration 
Ex - 
```
app.post('/book', (req, res) => {
   var email = res.email
   var firstName = res.firstName
});
```


pandurangpatil:
@khemrajrathore I guess this PR is no more in the context. Should we close this PR?

PR #200, opened 2022-11-17
Export the probable sinks to privado.json file

dbMundada:
- Currently we have limited set of rules written for Third Parties & Databases.
- So, after scanning it may be case that some Third Party is present but not detected via tool as rule missing.
- We will be exporting all the external packages in **probableSinks** key in privado.json

### DIFF
--- a/src/main/scala/ai/privado/exporter/SinkExporter.scala
+++ b/src/main/scala/ai/privado/exporter/SinkExporter.scala
@@ -29,16 +29,7 @@ import ai.privado.model.exporter.{SinkModel, SinkProcessingModel}
 import ai.privado.model.{CatLevelOne, Constants, DatabaseDetails, InternalTag}
 import ai.privado.semantic.Language.finder
 import io.shiftleft.codepropertygraph.generated.Cpg
-import io.shiftleft.codepropertygraph.generated.nodes.{
-  Call,
-  CfgNode,
-  FieldIdentifier,
-  Identifier,
-  Literal,
-  MethodParameterIn,
-  StoredNode,
-  Tag
-}
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, CfgNode, FieldIdentifier, Identifier, Literal, MethodParameterIn, StoredNode, Tag}
 import io.shiftleft.semanticcpg.language._
 import overflowdb.traversal.Traversal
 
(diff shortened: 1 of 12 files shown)


src/main/scala/ai/privado/exporter/SinkExporter.scala
```diff
@@ -55,8 +46,66 @@ class SinkExporter(cpg: Cpg) {
     convertSinkList(sinkTagList)
   }
 
-  def getProcessing: List[SinkProcessingModel] = {
+  def getProbableSinks: List[String] = {
+    var blackListedGroupIds: List[String] = List(
```
pandurangpatil, line 50: Convert this skiplist into config rule files picked from privado skiplist.yml configuration kept inside `config` folder.

pandurangpatil (changes requested):
handle the requested change.

src/main/scala/ai/privado/exporter/SinkExporter.scala
```diff
@@ -55,8 +46,66 @@ class SinkExporter(cpg: Cpg) {
...
+      "junit.framework.",
+      "io.reactivex.",
+      "com.bumptech.glide",
+      "com.google.protobuf",
+      "com.google.testing",
+      "com.google.common",
+      "<unresolvedNamespace>."
+    )
+    /** Get all the Methods which are tagged as SINKs  */
+    val taggedSinkMethods = cpg.tag.where(_.nameExact(Constants.catLevelOne).valueExact(CatLevelOne.SINKS.name)).call.l.map(i => i.methodFullName.split(":").head)
```
khemrajrathore, line 84: Instead of `head` use `headOption.getOrElse`. Head will return error if split return empty

src/main/scala/ai/privado/exporter/SinkExporter.scala
```diff
@@ -55,8 +46,66 @@ class SinkExporter(cpg: Cpg) {
...
+
+    /** Get all the Methods which are external  */
+    val dependenciesTPs = cpg.method.external.l.map(i => i.fullName.split(":").head)
```
khemrajrathore, line 87: Instead of `head` use `headOption.getOrElse`. Head will return error if split return empty

src/main/scala/ai/privado/exporter/SinkExporter.scala
khemrajrathore: Also `dedup` the list to get unique methodFullNames

src/main/scala/ai/privado/exporter/SinkExporter.scala
```diff
@@ -55,8 +46,66 @@ class SinkExporter(cpg: Cpg) {
...
+     * by excluding taggedSinkMethods
+     * using blackListedGroupIds
+     * transform method FullName close to groupIds
+     * remove duplicates
+     */
+    val filteredTPs = dependenciesTPs
+      .filter(str => !taggedSinkMethods.contains(str))
+      .filter((str) => {
+        val res = blackListedGroupIds.filter(blackListGroup => str.startsWith(blackListGroup))
+        (res.length == 0)
```
khemrajrathore, line 99: we can write this as `blackListedGroupIds.filter(blackListGroup => str.startsWith(blackListGroup)).empty`

src/main/scala/ai/privado/exporter/SinkExporter.scala
```diff
@@ -55,8 +46,66 @@ class SinkExporter(cpg: Cpg) {
...
+      })
+      .filter((str) => !str.endsWith(".println"))
+      .map((str) => str.split("\\.").take(3).mkString("."))
```
khemrajrathore, line 102: check if we need exception handling here

src/main/scala/ai/privado/exporter/SinkExporter.scala
khemrajrathore: @dbMundada

src/main/scala/ai/privado/exporter/SinkExporter.scala
khemrajrathore: @dbMundada

pandurangpatil (commented):
Looks good to me @khemrajrathore verify it once and if looks good from your side. Go ahead and merge it.

pandurangpatil (approved):

khemrajrathore (approved):
LGTM

PR #203, opened 2022-11-18
patch: update jackson-core version

abhstabs:
Jackson core issue: 

`com.networknt json-schema-validator` is using jackson-version `2.14.0`, causing issues with the other libraries. Upgrading version to 2.14.0 for fix.


pandurangpatil (approved):
Looks good to me.

PR #230, opened 2022-12-07
fix - picking correct rule in case the repo has JavaScript as the main language

khemrajrathore:
…n language


pandurangpatil (approved):
looks good to me.

PR #237, opened 2022-12-13
Alpha - Support python lang

tuxology:


khemrajrathore (dismissed):
@tuxology 
- Add new language `python` in PrivadoTag.scala --> Language.
- Also make appropriate change in ScanProcessor.scala --> parseRule --> langToFilter

tuxology:
Currently, because we don't have standard APIs for Python frontend, we get only base cpg with no overlays on it. We can handle running other overlays of course in Privado, but this is non-standard and will cause issue if upstream API changes eventually.  Therefore, I'll be working on this as a dependency for continuing work here: https://github.com/joernio/joern/issues/2113

build.sbt
```diff
@@ -22,7 +22,8 @@ dependsOn(domainClasses)
 libraryDependencies ++= Seq(
   "com.github.pathikrit"            %% "better-files"            % "3.9.1",
   "com.github.scopt"                %% "scopt"                   % "3.7.1",
-  "org.apache.logging.log4j"         % "log4j-slf4j-impl"        % "2.17.1"       % Runtime,
+  "org.apache.logging.log4j"         % "log4j-slf4j-impl"        % "2.18.0"       % Runtime,
+  "org.slf4j" % "slf4j-simple" % "2.0.6",
```
khemrajrathore, line 26: Run scalafmt

hiteshbedre (approved):

PR #257, opened 2022-12-22
Add sample types in snippet highlighted line

dbMundada:
We added functionality to show the type of the sample which getting highlighted in the code snippet.

image


pandurangpatil:
@khemrajrathore  can you review this Pull request and merge it to dev if looks good to you? We need to check if we are adding new fields or modifying existing fields. If we are adding new field, that should not create issues. If we are removing an existing field or changing any existing field then we need to see how we can keep it backwards compatible.

dbMundada:
This PR will only impact those who are processing code-snippets within the flows for some purpose.
@pandurangpatil @khemrajrathore

khemrajrathore (approved):
LGTM

PR #294, opened 2023-01-24
Optimized rules parsing using parallelization

karan-batavia:


src/main/scala/ai/privado/model/PrivadoTag.scala
```diff
@@ -87,6 +87,7 @@ object Language extends Enumeration {
 
   val JAVA       = Value("java")
   val JAVASCRIPT = Value("javascript")
+  val PYTHON = Value("python")
   val DEFAULT    = Value("default")
```
khemrajrathore, line 91: remove the ImportUtility related implementation, that will go as part of enhanced API branch

PR #296, opened 2023-01-24
Fix/exclude test file in property pass

khemrajrathore:
fix - To not consider `src/test` file in property file pass
add - To only consider properties in `resources` folder as properties inside `WEB-INF` etc also gets considered currently


pandurangpatil (approved):
Looks good to me.

PR #306, opened 2023-01-31
feature: add support for system config files

abhstabs:


src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
```diff
@@ -99,8 +100,14 @@ class JavaAPITagger(cpg: Cpg) extends ForkJoinParallelCpgPass[RuleInfo](cpg) {
   }
 
   implicit val engineContext: EngineContext = EngineContext(semantics = getDefaultSemantics, config = EngineConfig(4))
-  val commonHttpPackages: String =
-    "^(?i)(org.apache.http|okhttp|org.glassfish.jersey|com.mashape.unirest|java.net.http|java.net.URL|org.springframework.(web|core.io)|groovyx.net.http|org.asynchttpclient|kong.unirest.java|org.concordion.cubano.driver.http|javax.net.ssl).*"
+  val systemConfigHttpLibraries = RuleCache.getRule.systemConfig.filter(config => config.key.equals("apiHttpLibraries"))
```
khemrajrathore, line 103: @abhstabs 
- What happens if older Privado is passed. (is this backward compatible)?
- If 2 privado are passes, one and internal and one as external both having the same key, what is the behaviour?

Also better to create a wrapper function in RuleCache where we just say `RuleCache.getSystemConfigById(id)` and it returns a value, so that it is easy to use

khemrajrathore: @abhstabs We have few more regex used in API tagger, if we can move them in config as well

abhstabs: 1. Backwards compatible. If no apiHttpLibraries are passed, we use a default set of values. 
2. If 2 or more configs have the same value, or even in case of same keys having different value, the result will be a combined regex of all the values.

khemrajrathore (approved):
LGTM

PR #333, opened 2023-02-09
Python update probable sinks

dbMundada:


khemrajrathore (commented):
Generate comparison result with Upgraded Version so that there is no delta, Rest looks good

khemrajrathore:
@dbMundada once we have result we can take this forward

khemrajrathore (approved):
LGTM

PR #337, opened 2023-02-10
Bugfix/rm warnings

gaurav-gogia:
- rm duplicate function from `Utilities.scala`
- use string interpolation in unresolved function reporter
- use `CpgPass` over depreciated `SimpleCpgPass`


src/main/scala/ai/privado/tagger/PrivadoSimplePass.scala
```diff
@@ -25,10 +25,10 @@ package ai.privado.tagger
 import ai.privado.metric.MetricHandler
 import ai.privado.model.RuleInfo
 import io.shiftleft.codepropertygraph.generated.Cpg
-import io.shiftleft.passes.SimpleCpgPass
+import io.shiftleft.passes.CpgPass
 import org.slf4j.LoggerFactory
 
-abstract class PrivadoSimplePass(cpg: Cpg) extends SimpleCpgPass(cpg) {
```
abhstabs, line 31: @pandurangpatil Can you confirm if this is correct? Using CpgPass instead of SimpleCpgPass.

pandurangpatil: @abhstabs is right, keep it as SimpleCpgPass for now. We need to test the changes for this one separately. We will create a separate ticket for this one.

src/main/scala/ai/privado/utility/Utilities.scala
```diff
@@ -424,117 +423,4 @@ object Utilities {
     })
     semanticList.toList
   }
-
-  def reportUnresolvedMethods(xtocpg: Try[Cpg], filename: String): Unit = {
```
pandurangpatil, line 428: @gaurav-gogia did we remove this method?

pandurangpatil (changes requested):
Need to revert back to SimpleCpgPass for now.

src/main/scala/ai/privado/utility/Utilities.scala
gaurav-gogia: @pandurangpatil, this one is a duplicate. I forgot to remove it before by mistake. Correct one can be found here:
`src/main/scala/ai/privado/utility/UnresolvedReportUtility.scala`

pandurangpatil (approved):
Looks good

PR #350, opened 2023-02-14
Fix/single node multiple sink exporting

khemrajrathore:


src/main/scala/ai/privado/dataflow/DuplicateFlowProcessor.scala
```diff
@@ -313,13 +312,15 @@ object DuplicateFlowProcessor {
...
         }
-        if (ScanProcessor.config.disableFlowSeparationByDataElement)
-          DataFlowCache.setDataflow(
-            DataFlowPathModel(pathSourceId, sinkId, dataflowSinkType, dataflowNodeType, sinkPathId)
-          )
-        else
-          filterFlowsOverlappingWithOtherDataElement()
-        AppCache.totalMap.put(dataflowSinkType, AppCache.totalMap.getOrElse(dataflowSinkType, 0) + 1)
+        sinkCatLevelTwoCustomTag.value.foreach(sinkId => {
+          if (ScanProcessor.config.disableFlowSeparationByDataElement)
```
pandurangpatil, line 316: Is this and improvement in false positive?

khemrajrathore: No, this is regarding a single sink node getting tagged with multiple sinks, in that case, we need to show the same flow across Sinks

pandurangpatil (approved):
Looks good to me.

PR #355, opened 2023-02-15
Refactoring parameter 

ankit-privado:
Changed the parameter name and create directory if not exist for intermediate result.


pandurangpatil:
We also need to refactor the code to generate the output for JavaScript and Python repos. With current call flow it will only be generated for Java.

ankit-privado:
@pandurangpatil yes sure will make the change for javascript and python. this change is related to generation for report.

pandurangpatil (approved):
Looks good

PR #379, opened 2023-03-01
added collection support on literal

hiteshbedre:


src/main/scala/ai/privado/exporter/CollectionExporter.scala
```diff
@@ -101,6 +107,29 @@ class CollectionExporter(cpg: Cpg) {
         })
     })
 
+    collectionMethods.foreach(collectionMethod => {
+      collectionMethod.literal
+        .and(_.tag.nameExact(Constants.id))
+        .foreach(literal => {
+          try {
+            def addToMap(literalId: String): Unit = {
```
khemrajrathore, line 115: @hiteshbedre I see the function `addToMap` getting repeated 3 times, better to make a generic one and use

src/main/scala/ai/privado/exporter/CollectionExporter.scala
```diff
@@ -167,6 +198,23 @@ class CollectionExporter(cpg: Cpg) {
     )
   }
 
+  def processByLiteralId(
```
khemrajrathore, line 201: @hiteshbedre Also check if possible to have a single function processById, instead of `processByParameterId`, `processByLocalVariableId` and `processByLiteralId`

src/main/scala/ai/privado/languageEngine/python/tagger/collection/CollectionTagger.scala
```diff
@@ -63,13 +63,15 @@ class CollectionTagger(cpg: Cpg, sourceRuleInfos: List[RuleInfo]) extends ForkJo
         // TODO: handle cases where `request.args.get('id', None)` used directly in handler block without method param
         val parameters =
           collectionMethod.parameter.where(_.name(sourceRule.combinedRulePattern)).whereNot(_.code("self")).l
-        val matchingLocals = collectionMethod.local.code(sourceRule.combinedRulePattern).l
+        val matchingLocals   = collectionMethod.local.code(sourceRule.combinedRulePattern).l
+        val matchingLiterals = collectionMethod.literal.code(sourceRule.combinedRulePattern).l
```
khemrajrathore, line 67: @hiteshbedre Don't match literal's like this, it will result in a lot of FP. Refer the literalTagger in Java.
We can tag literals only when they are argument of the `get` kind of function calls

khemrajrathore (commented):
@hiteshbedre One question, If a Collection point has match for both local and literal, is it getting repeated twice?

hiteshbedre:
Screenshot 2023-03-01 at 6 06 29 PM
Yes, it will repeated twice but shown under same url.

khemrajrathore:
@hiteshbedre Would recommend to add some intelligence like if already a method is identified via local for a data element not to show it for literal

src/main/scala/ai/privado/exporter/CollectionExporter.scala
hiteshbedre: Handled here: https://github.com/Privado-Inc/privado-core/pull/379/commits/49e4553a24590a35a56d5d319be1b0bd81f84b99

src/main/scala/ai/privado/languageEngine/python/tagger/collection/CollectionTagger.scala
hiteshbedre: @khemrajrathore This was really really helpful. Can able to see lots of false positive literals can be eliminated with this well handled query🙌🏻
Commit: https://github.com/Privado-Inc/privado-core/pull/379/commits/487a37602f827be6fc28a9d76c0e58afd1e33011

src/main/scala/ai/privado/exporter/CollectionExporter.scala
hiteshbedre: Need a parent class which has properties of AstNode also supports _.method_ method over it, not found as of now. Will spend some extra time on this soon.

src/main/scala/ai/privado/languageEngine/python/tagger/collection/CollectionTagger.scala
```diff
@@ -64,12 +64,20 @@ class CollectionTagger(cpg: Cpg, sourceRuleInfos: List[RuleInfo]) extends ForkJo
         val parameters =
           collectionMethod.parameter.where(_.name(sourceRule.combinedRulePattern)).whereNot(_.code("self")).l
         val matchingLocals = collectionMethod.local.code(sourceRule.combinedRulePattern).l
-
-        if (parameters.isEmpty && matchingLocals.isEmpty) {
+        val matchingLiterals = collectionMethod
+          .call("(?:get).*")
+          .argument
+          .isLiteral
+          .code("\"(" + sourceRule.combinedRulePattern + ")\"")
```
khemrajrathore, line 71: @hiteshbedre along with `"` also include `'` as single and double quotes both can be used for literal in python

hiteshbedre: Handled here: https://github.com/Privado-Inc/privado-core/pull/379/commits/9b7f2b2eeebe69aad75f39683db4f09ad679e32e

PR #390, opened 2023-03-06
Cpg location

karan-batavia:


src/main/scala/ai/privado/languageEngine/python/processor/PythonProcessor.scala
```diff
@@ -30,6 +30,7 @@ import ai.privado.model.Language
 import java.util.Calendar
 import scala.jdk.CollectionConverters.CollectionHasAsScala
 import scala.util.{Failure, Success, Try}
+import java.nio.file.{Paths, Files}
 
 object PythonProcessor {
   private val logger = LoggerFactory.getLogger(getClass)
```
pandurangpatil, line 36: Change the file location to use the new `cpg.bin` file location at line no 122. Following is the code snippet you need to change
```
121          import java.io.File
122          val cpgFile = new File(sourceRepoLocation)
123          println(s"\n\nBinary file size -- ${cpgFile.length()} in Bytes - ${cpgFile.length() * 0.000001} MB\n")
```

pandurangpatil (changes requested):
Made respective change for PythonProcessor.

src/main/scala/ai/privado/languageEngine/java/processor/JavaProcessor.scala
```diff
@@ -158,7 +159,15 @@ object JavaProcessor {
       println(s"${Calendar.getInstance().getTime} - Downloading dependencies and Parsing source code...")
     else
       println(s"${Calendar.getInstance().getTime} - Parsing source code...")
-    cpgconfig = Config(inputPath = sourceRepoLocation, fetchDependencies = !config.skipDownloadDependencies)
+
+
+
+    if (!Files.exists(Paths.get("./.privado"))) {
```
khemrajrathore, line 165: @karan-batavia As this is a repeated function across, expose this as a function and use the function wherever required. Also test out this implementation in docker version

src/main/scala/ai/privado/languageEngine/java/processor/JavaProcessor.scala
```diff
@@ -158,7 +159,15 @@ object JavaProcessor {
...
+      Files.createDirectory(Paths.get("./.privado"));
+    }
+
+    val cpgOutputPath = s"${Paths.get(".").toAbsolutePath}/.privado/cpg.bin"
```
khemrajrathore, line 169: @karan-batavia the name `.privado` is already part of Constants, use that so that in future we just have to make change in the constants to change the output folder. Also the name of the binary which is `cpg.bin` store that in Constants as well

pandurangpatil:
@khemrajrathore we will test this change with the docker build once it's been merged to `dev`. It's not worth spending time to set up a separate build for testing this one with the docker build.

pandurangpatil (approved):
looks good to me.

PR #391, opened 2023-03-06
fix - dedup on api flows

khemrajrathore:
Solving the problem were `slackWebHookUrl` point to the URL `hooks.slack.com` and we show 2 API's

The flow of slackWebHookUrl is a sub-flow of the flow starting from hooks.slack.com and hence will be discarded by deduplication logic


pandurangpatil (approved):
looks good

PR #400, opened 2023-03-08
Feign

khemrajrathore:
- Refactoring to place Feign outside of JavaAPITagger
- More support for Url detection for Feign


pandurangpatil (approved):
looks good to me

PR #402, opened 2023-03-09
Add separate LiteralTagger for python with sql literal supports

dbMundada:


src/main/scala/ai/privado/languageEngine/python/tagger/source/LiteralTagger.scala
```diff
@@ -0,0 +1,63 @@
...
+  override def generateParts(): Array[RuleInfo] = RuleCache.getRule.sources.toArray
+  override def runOnPart(builder: DiffGraphBuilder, ruleInfo: RuleInfo): Unit = {
+    // Step 1.2
+    // val literals = cpg.literal.code("\"(" + ruleInfo.patterns.head + ")\"").whereNot(_.code(".*\\s.*")).l
+    val generalLiterals = cpg
+      .call("(?:add|get|put|pop).*")
+      .argument
+      .where(_.argumentIndex(1))
+      .isLiteral
+      .code("\"(" + ruleInfo.combinedRulePattern + ")\"")
```
khemrajrathore, line 44: @dbMundada Also include single quote as string in python can be defined with single quote

khemrajrathore (approved):
LGTM

PR #404, opened 2023-03-09
add - consider members for identifier regex in API

khemrajrathore:
- Consider Member for API identifier Regex
- Consider Identifier Regex for Feign as well


pandurangpatil (approved):
looks good to me.

PR #413, opened 2023-03-13
LE-242: Used null when the list is blank and associated null check

rahul-privado:
### Description

The exception is due to dereferencing the first element of a blank list. Example of code that causes this is [here](https://github.com/hiteshbedre/Finance-System/blob/4c855dd13b820f49d151ac77b99851a9d421c33f/accounts/src/main/java/com/bhanuchaddha/bank/accounts/Account.java#L21).

### Fix
Put a safety null check

### Testing done
Tested on the repo mentioned in the ticket. Also tested on `privado-accounts-api`

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePass.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePass.scala
@@ -93,6 +93,7 @@ class PropertiesFilePass(cpg: Cpg, projectRoot: String) extends ForkJoinParallel
     val membersAndValues = annotatedMembers()
 
     membersAndValues
+      .filter(_._1 != null)
       .filter { case (key, _) => propertyNode.name == key.code.slice(3, key.code.length - 2) }
       .foreach { case (_, value) =>
         builder.addEdge(propertyNode, value, EdgeTypes.IS_USED_AT)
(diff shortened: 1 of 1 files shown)


src/main/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePass.scala
```diff
@@ -118,7 +119,13 @@ class PropertiesFilePass(cpg: Cpg, projectRoot: String) extends ForkJoinParallel
   private def annotatedMembers() = cpg.annotation
     .fullName(".*Value.*")
     .where(_.member)
-    .map { x => (x.parameterAssign.head, x.member.head) }
+    .map { x =>
+      {
+        val param = if (x.parameterAssign.l.length > 0) { x.parameterAssign.head }
+        else { null }
```
khemrajrathore, line 125: @rahul-privado instead of introduction `null`, you can try converting the `map` to  a `flatMap` and
```
if (x.parameterAssign.l.length >0)
         Some(param, x.member.head)
else 
         None

rahul-privado: Used a filter and just checked in. Please check

khemrajrathore: Also earlier as I said `cpg.annotation.fullName(".*Value.*")` is not the correct way, it can match any annotation having `.*Value.*` in it, instead of Springframework's Value annotation.

rahul-privado: I have tried `.fullName("org.springframework.*Value")` but that does not work. Can you point me to the slack link where you have raised it with the joern team? I will follow up.

khemrajrathore (approved):
LGTM

khemrajrathore (approved):
LGTM

PR #417, opened 2023-03-14
Patch inherit semantic

khemrajrathore:
```
@Getter
class Abstract BaseClass {
private String id;
private String firstName;
}

@Getter
class Payment {
private String paymentMethod;
}


log.info(payment.getId());
```
 

Payment class here has only 1 member which is `paymentMethod` but `id` is not a member of `Payment` class, but it happens to extend `BaseClass` which has a member `id` and the function `getId` is now exposed to `Payment` as it has inherited `BaseClass`. So we were not able to generate a semantic for this case as `id` was not a member of Payment.


pandurangpatil (approved):
looks good to me

PR #421, opened 2023-03-16
GRPC collections pass

tuxology:
This pass aims to model GRPC server request handlets like this as "Collections":

```java
    public void sayHello(HelloRequest req, StreamObserver<HelloReply> responseObserver) {
      // get client id added to context by interceptor
      String clientId = Constant.CLIENT_ID_CONTEXT_KEY.get();
      logger.info("Processing request from " + clientId);
      HelloReply reply = HelloReply.newBuilder().setMessage("Hello, " + req.getName()).build();
      responseObserver.onNext(reply);
      responseObserver.onCompleted();
    }
```
https://github.com/grpc/grpc-java/blob/master/examples/example-jwt-auth/src/main/java/io/grpc/examples/jwtauth/AuthServer.java#L95-L102 

Currently we are not considering if `req` contains any sensitive data or not. But will explore later. The pass is able to tag the collections correctly, but nothing is shown in UI possibly because the way UI works is different than our simple approach. 

**Testing**

Changes have been tested on 6 repos. 3 gRPC, 3 non-gRPC. Non-gRPC repos were scanned with released privado and this branch to compare results. Have not found any problematic delta between those scans.

GRPC Repos
- https://github.com/anjeyy/grpc-sample
- https://github.com/tkvangorder/sleuth-grpc-sample
- https://github.com/revinate/grpc-spring-security-demo

Non-GRPC Repos
- https://github.com/saurabh-sudo/BankingSystem-Backend
- https://github.com/webtutsplus/ecommerce-backend
- https://github.com/antkaynak/Keyist-Ecommerce


src/main/scala/ai/privado/languageEngine/java/tagger/collection/GrpcCollectionTagger.scala
```diff
@@ -0,0 +1,49 @@
...
+import io.shiftleft.semanticcpg.language._
+import org.slf4j.LoggerFactory
+
+class GrpcCollectionTagger(cpg: Cpg) extends CpgPass(cpg) {
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  override def run(builder: DiffGraphBuilder): Unit = {
+
+    val grpcCollectionPoints = cpg.method
+      .where(_.call.name("onCompleted")) // grpc server always call this in handler when grpc-java is used
```
gaurav-gogia, line 39: I think we can qualify this `onCompleted` method further by adding number of params and `StreamObserver` in methodFullName filter

`val OnCompletedPattern = "(?i)(.*)(StreamObserver)(.*)(onCompleted)(.*)([(](0)[)])$"`

tuxology: Maybe instead of using regex for param count, we can check via argument index on the call. Let me see if it's available.

gaurav-gogia (approved):

src/main/scala/ai/privado/languageEngine/java/tagger/PrivadoTagger.scala
```diff
@@ -103,6 +103,16 @@ class PrivadoTagger(cpg: Cpg) extends PrivadoBaseTagger {
...
     )
+
+    println(s"${Calendar.getInstance().getTime} - --GrpcCollectionTagger invoked...")
+    new GrpcCollectionTagger(cpg, sourceRules).createAndApply()
+    println(
+      s"${TimeMetric.getNewTime()} - --GrpcCollectionTagger is done in \t\t\t- ${TimeMetric.setNewTimeToStageLastAndGetTimeDiff()}"
+    )
+
+    // TODO: Remove after debugging
+    println(cpg.method.where(_.call.name("onCompleted")).tag.toJsonPretty)
```
khemrajrathore, line 114: remove print

khemrajrathore (commented):
Here we are only tagging Collection Points whose parameter's match the source Rule. 
Ex - If parameter is `firstName`, `lastName`

don't we also want to tag derived sources like `user` which has `firstName` as a member?

khemrajrathore (approved):
LGTM

PR #424, opened 2023-03-18
Dbconfig/neo4j

HiteshMah-Jan:


src/main/scala/ai/privado/tagger/config/DBConfigTagger.scala
```diff
@@ -55,6 +55,10 @@ class DBConfigTagger(cpg: Cpg) extends ForkJoinParallelCpgPass[JavaProperty](cpg
         parsePropForSpringJDBCAndJPA(dbUrl)
       } else if (dbUrl.value.contains("mongodb")) {
         parsePropForSpringDataMongo(dbUrl)
+      } else if (dbUrl.name.contains("neo4j.host") && dbUrl.value.matches("(localhost|[^{]*\\..*\\.[^}]*)")) {
```
khemrajrathore, line 58: @HiteshMah-Jan Add comment regarding what are we looking to match with the below regex

khemrajrathore (approved):
LGTM

PR #425, opened 2023-03-20
segregation of FP flag between language

khemrajrathore:
- Caching the language the repo contains, so that we can use the cache latter
- Adding a condition, so that the FP flags `disable-this-filtering` and `disable-segregation-by-dataelement` is only done for JAVA language, and it doesn't filter out flows in other Languages


pandurangpatil (approved):
looks good

PR #439, opened 2023-03-23
Python db config

karan-batavia:


src/main/scala/ai/privado/languageEngine/python/passes/config/PythonPropertyFilePass.scala
```diff
@@ -0,0 +1,178 @@
...
+      if (d.exists && d.isDirectory) {
+        d.listFiles.filter(_.isFile).toList
+      } else {
+        List[File]()
+      }
+    }
+
+    SourceFiles
+      .determine(Set(projectRoot), extensions)
+      .concat(getListOfFiles(projectRoot).map(f => f.getAbsolutePath).filter(_.matches(".*\\.env.*")))
```
khemrajrathore, line 103: @karan-batavia Why did you removed `.filter(Utilities.isFileProcessable)` , I asked to add this to the output of getListOfFiles so that we filter invalid property files

khemrajrathore (approved):
LGTM

PR #450, opened 2023-03-27
Added unit test check support in workflow

ankit-privado:
Added Unit Test check support in current workflow and if any of the unit tests fail, the entire workflow will stop.

Ticket: https://privado-ai.atlassian.net/browse/LE-293


pandurangpatil (approved):
looks good

PR #463, opened 2023-03-31
ead patch

gaurav-gogia:
Patching API sink call tagging in V1 so that `-ead` flag does not have to be used for gRPC sinks


khemrajrathore (approved):
LGTM

PR #481, opened 2023-04-04
Update/free grpc sinks

gaurav-gogia:
**Features**
- Add Independent tagging of grpc sink calls that does **not** rely on collection tagging
- Imrove `onNext` function detection to fix FPs in gRPC sink call detection

- Tested on `grpc-sample` repo so far
- This implementation is not perfect, may lead to some false negatives
- More intensive testing is required to qualify sink calls better

Testing repo links:
- https://github.com/anjeyy/grpc-sample

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/Utility/GRPCTaggerUtility.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/Utility/GRPCTaggerUtility.scala
@@ -30,10 +30,11 @@ object GRPCTaggerUtility {
   def getGrpcSinkCalls(cpg: Cpg, grpcEndpoints: List[Method]): List[Call] = {
     // `onNext` is always called inside of gRPC service methods
     // This `onNext` always has a StreamObserver in signature and takes only one argument
-    val onNext        = "onNext"
-    val stubPattern   = "(?i)(.*)(stub)(.*)"
-    val anyType       = "ANY"
-    var grpcSinkCalls = ListBuffer[Call]()
+    val onNext           = "onNext"
+    val stubPattern      = "(?i)(.*)(stub)(.*)"
+    val anyType          = "ANY"
+    val filterNotPattern = "(?i)(.*)(operator)(.*)"
+    var grpcSinkCalls    = ListBuffer[Call]()
 
     grpcEndpoints.foreach(endpoint => {
       // Detecting gRPC API sink calls. These sink calls have the same name as function calls inside of onNext functions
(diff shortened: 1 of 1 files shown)


khemrajrathore (commented):
@gaurav-gogia Should we trigger GRPC tagger only when we can confirm if a repository has GRPC implementation. The idea to not run GRPC thing for nonGRPC repos, the queries you have used a abstract, so better to run it only on GRPC repos. Can we thing of some condition to identify if a repo has a GRPC implementation?

gaurav-gogia:
> @gaurav-gogia Should we trigger GRPC tagger only when we can confirm if a repository has GRPC implementation. The idea to not run GRPC thing for nonGRPC repos, the queries you have used are abstract, so better to run it only on GRPC repos. Can we thing of some condition to identify if a repo has a GRPC implementation?

I'm onboard with your idea of calling gRPC scan only when it is required. But detecting if a repo is using gRPC or not could get tricky.

Well, we can try two methods here to detect if a repo is using gRPC

- Look for proto files in the same repo
- Look for GrpcService annotation
- Look for protobuf/gRPC in imports
- Look for protobuf/gRPC in pom.xml 

I'll spend some time on it to figure out the best approach. 

How urgent is it though? Can we do it in a future PR?

khemrajrathore:
> > @gaurav-gogia Should we trigger GRPC tagger only when we can confirm if a repository has GRPC implementation. The idea to not run GRPC thing for nonGRPC repos, the queries you have used are abstract, so better to run it only on GRPC repos. Can we thing of some condition to identify if a repo has a GRPC implementation?
> 
> I'm onboard with your idea of calling gRPC scan only when it is required. But detecting if a repo is using gRPC or not could get tricky.
> 
> Well, we can try two methods here to detect if a repo is using gRPC
> 
> * Look for proto files in the same repo
> * Look for GrpcService annotation
> * Look for protobuf/gRPC in imports
> * Look for protobuf/gRPC in pom.xml
> 
> I'll spend some time on it to figure out the best approach.
> 
> How urgent is it though? Can we do it in a future PR?

The earlier the better, otherwise we will see False Positives for GRPC which will degrade the result. Will approve this for now, but work on this

khemrajrathore (approved):
LGTM

gaurav-gogia:
> > > @gaurav-gogia Should we trigger GRPC tagger only when we can confirm if a repository has GRPC implementation. The idea to not run GRPC thing for nonGRPC repos, the queries you have used are abstract, so better to run it only on GRPC repos. Can we thing of some condition to identify if a repo has a GRPC implementation?
> > 
> > 
> > I'm onboard with your idea of calling gRPC scan only when it is required. But detecting if a repo is using gRPC or not could get tricky.
> > Well, we can try two methods here to detect if a repo is using gRPC
> > 
> > * Look for proto files in the same repo
> > * Look for GrpcService annotation
> > * Look for protobuf/gRPC in imports
> > * Look for protobuf/gRPC in pom.xml
> > 
> > I'll spend some time on it to figure out the best approach.
> > How urgent is it though? Can we do it in a future PR?
> 
> The earlier the better, otherwise we will see False Positives for GRPC which will degrade the result. Will approve this for now, but work on this

Sure, will create a ticket to keep track of this. Will add it in next PR.

PR #483, opened 2023-04-05
Support: Sentry python

hiteshbedre:
A complete code on how to integrate Sentry:

  ```
 sentry_sdk.init(
    dsn="https://xyz@abc.ingest.sentry.io/89635676",

    # Set traces_sample_rate to 1.0 to capture 100%
    # of transactions for performance monitoring.
    # We recommend adjusting this value in production.
    traces_sample_rate=1.0
)


import logging
logging.error("I am an event", extra=dict(bar=43))
logging.exception("An exception happened")

```

Whenever error or exception log gets executed, a record will passed on sentry.io
We need to capture such statement if PII is going out from the system.


khemrajrathore (approved):
LGTM

PR #479, opened 2023-04-03
add - soap api and collection point where method is denoted by webser…

khemrajrathore:
add - soap api and collection point where method is denoted by webservice annotation

```
@WebService(name = "ContourSoapService", targetNamespace = "http://v3.ws.contour.jamasoftware.com/")
@XmlSeeAlso({
        ObjectFactory.class
})
public interface ContourSoapService {


    /**
     * @param organization
     * @param token
     * @return returns com.jamasoftware.examples.soap.wsdl.WsOrganization
     */
    @WebMethod(action = "updateOrganization")
    @WebResult(targetNamespace = "")
    @RequestWrapper(localName = "updateOrganization", targetNamespace = "http://v3.ws.contour.jamasoftware.com/", className = "com.jamasoftware.examples.soap.wsdl.UpdateOrganization")
    @ResponseWrapper(localName = "updateOrganizationResponse", targetNamespace = "http://v3.ws.contour.jamasoftware.com/", className = "com.jamasoftware.examples.soap.wsdl.UpdateOrganizationResponse")
    WsOrganization updateOrganization(
            @WebParam(name = "token", targetNamespace = "")
            WsAuth token,
            @WebParam(name = "organization", targetNamespace = "")
            WsOrganization organization);
```

PR #488, opened 2023-04-06
add - soap api and collection point where method is denoted by webser…

khemrajrathore:
…… (#479)

* add - soap api and collection point where method is denoted by webservice annotation

* refactoring and test case

* more test cases

* more testcase

* refactoring

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Privado Core
 =============================================
 
-Branch structure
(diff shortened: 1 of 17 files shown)


pandurangpatil (approved):
looks good to me

pandurangpatil (approved):
looks good

PR #507, opened 2023-04-17
added tortoise support

hiteshbedre:
Change:

1. Added custom inherit tagger for Python
2. Added storage rule for Tortoise


hiteshbedre:
Removed external rule for tortoise here: https://github.com/Privado-Inc/privado/pull/226

src/main/scala/ai/privado/languageEngine/java/feeder/StorageInheritRule.scala
```diff
@@ -172,6 +172,38 @@ object StorageInheritRule {
       "storages",
       Language.JAVA,
       Array[String]()
+    ),
```
khemrajrathore, line 175: Why are we storing python rules in Java?

PR #520, opened 2023-04-21
PrivadoPythonTypeHintCallLinker: Preserve Old Type Hint Call Linker Behaviour

DavidBakerEffendi:
Introduce `PrivadoPythonTypeHintCallLinker` to preserve old call linker behaviour and not lose current flows until the next fix in Joern for scope presence of symbol table values.


pandurangpatil (approved):
looks good

PR #527, opened 2023-04-26
Fix for comparison report diff

pandurangpatil:
HtmlUnit library is not threadsafe. It in itself make use of multiple threads for processing single HTML file. Because we tried to use same instance across multiple threads. It is resulting into inconsistent outcome. Hence changed it to be instanciated for each thread separately.

Also added trace logs.


khemrajrathore (approved):
LGTM

PR #534, opened 2023-04-30
Added Dependency Category support

ankit-privado:
Added dependency category, process, rule in dependency sheet
Ticket: https://privado-ai.atlassian.net/browse/LE-274


src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -173,7 +176,37 @@ object ScanProcessor extends CommandProcessor {
                             language = Language.withNameWithDefault(pathTree.last)
                           )
                         )
-                        .filter(filterSystemConfigByLang)
+                        .filter(filterSystemConfigByLang),
+                      auditCollection = configAndRules.auditCollection
```
khemrajrathore, line 180: What are these 3 new rules and will this be part of Privado repo?

ankit-privado: There three new rule is for audit dependency only. so to identify which dependency belong to which category (Collection, Web client, Utility) we need rules. Yes, we also need to define the rule inside Privado repo.

khemrajrathore:
Changes good to Me, but will wait for the PR of Privado

pandurangpatil (approved):
looks good

PR #550, opened 2023-05-05
fixed regex for env files in property file pass

karan-batavia:


src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -312,7 +312,13 @@ class PropertyParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache, la
 
     SourceFiles
       .determine(Set(projectRoot), extensions)
-      .concat(getListOfFiles(projectRoot).map(f => f.getAbsolutePath).filter(_.matches(".*\\.env.*")))
+      .concat(
+        getListOfFiles(projectRoot)
+          .map(f => {
+            f.getAbsolutePath
+          })
+          .filter(_.matches(".*\\.env(?!.*(?:.js|.py|.java|.sh)$).*"))
```
dbMundada, line 320: Add .json, .ts also in filter

karan-batavia: We are scanning .json files separately. Will add .ts

dbMundada (approved):

PR #552, opened 2023-05-08
Fix for Property File Pass failing

karan-batavia:


dbMundada (approved):
LGTM

src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -128,19 +128,23 @@ class PropertyParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache, la
...
-        } catch {
-          case e: Throwable =>
-            logger.debug(s"Error splitting the required line. ${e.toString}")
-        }
-      })
+    try {
+      Source
+        .fromFile(file)
+        .getLines()
+        .filter(line => line.trim.nonEmpty && !line.startsWith("#"))
```
khemrajrathore, line 135: What happens for multi line comments?

khemrajrathore: How, If there are configs which are commented, nodes will get created for them

karan-batavia: Ah, this might be a case. But as long as I can remember,  there are no multiline comments in .env files. Let me check again and add support if that's not the case.

PR #553, opened 2023-05-09
YML/YAML Support added for python property parser

karan-batavia:


src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -53,7 +53,10 @@ class PropertyParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache, la
       }
       case Language.JAVASCRIPT => configFiles(projectRoot, Set(FileExtensions.JSON)).toArray
       case Language.PYTHON =>
-        configFiles(projectRoot, Set(FileExtensions.INI, FileExtensions.ENV, FileExtensions.CONF)).toArray
+        configFiles(
+          projectRoot,
+          Set(FileExtensions.INI, FileExtensions.ENV, FileExtensions.YAML, FileExtensions.YML)
```
dbMundada, line 58: CONF missing, is it expected?

karan-batavia: Yes, We're not scanning CONF for python.

PR #554, opened 2023-05-10
Export run time semantics to a file

karan-batavia:
Exporting run-time semantics to the .privado folder

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/semantic/JavaSemanticGenerator.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/semantic/JavaSemanticGenerator.scala
@@ -27,6 +27,7 @@ import ai.privado.cache.RuleCache
 import ai.privado.entrypoint.PrivadoInput
 import ai.privado.model.{CatLevelOne, Constants, InternalTag}
 import ai.privado.semantic.SemanticGenerator
+import ai.privado.utility.Utilities.semanticExporter
 import io.joern.dataflowengineoss.semanticsloader.{Parser, Semantics}
 import io.shiftleft.codepropertygraph.generated.Cpg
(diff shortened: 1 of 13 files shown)


src/main/scala/ai/privado/languageEngine/java/semantic/JavaSemanticGenerator.scala
```diff
@@ -42,7 +43,7 @@ object JavaSemanticGenerator extends SemanticGenerator {
     *   \- cpg for adding customSemantics
     * @return
     */
-  def getSemantics(cpg: Cpg, privadoScanConfig: PrivadoInput, ruleCache: RuleCache): Semantics = {
+  def getSemantics(cpg: Cpg, privadoScanConfig: PrivadoInput, ruleCache: RuleCache, repoPath: String): Semantics = {
```
khemrajrathore, line 46: The repoPath as per my understanding is available in `privadoScanConfig`

src/main/scala/ai/privado/utility/Utilities.scala
```diff
@@ -371,4 +373,24 @@ object Utilities {
     builder.addNode(node)
     builder.addEdge(node, fileNode, EdgeTypes.SOURCE_FILE)
   }
+
+  def semanticExporter(sourceRepoLocation: String, headers: List[String], semantics: List[Seq[String]]): Unit = {
```
khemrajrathore, line 377: Rename this to semanticFileExporter

khemrajrathore, line 377: Instead of taking 2 lists as input, better take a map as input with key being the Header and value Seq/List of String

khemrajrathore (approved):
LGTM

PR #563, opened 2023-05-18
Python yml

karan-batavia:


src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -155,30 +160,37 @@ class PropertyParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache, la
...
+        case sequenceNode: SequenceNode =>
+          sequenceNode.getValue.asScala.zipWithIndex.foreach { case (valueNode, index) =>
+            val fullPath = s"$path[$index]"
+            processNode(valueNode, fullPath)
+          }
+        case scalarNode: ScalarNode =>
+          val line   = scalarNode.getStartMark.getLine + 1
+          val column = scalarNode.getStartMark.getColumn + 1
+          val value  = scalarNode.getValue
+          result = result.appended((path, s"$value...$line...$column"))
```
khemrajrathore, line 189: instead of concating with `...` and splitting again aftwards, can we use a tuple or some Data Structure?

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
```diff
@@ -0,0 +1,100 @@
...
+    cpg.property.l.filter(prop => prop.name.matches("(?i)query")).toArray
+
+  override def runOnPart(builder: DiffGraphBuilder, property: JavaProperty): Unit = {
+    addSqlNodes(builder, property)
+  }
+
+  private def addSqlNodes(builder: SQLPropertyPass.this.DiffGraphBuilder, property: JavaProperty): Unit = {
+
+    val query        = property.value
+    val lineNumber   = property.lineNumber.getOrElse(-1).asInstanceOf[Int]
```
khemrajrathore, line 34: Instead of assingning -1 as reading time, better when we create Nodes in Property pass, we assign the lineNumber as -1 and columnNumber as -1, if we are unable to fetch the value

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
```diff
@@ -0,0 +1,100 @@
...
+    val columnNumber = property.columnNumber.getOrElse(-1).asInstanceOf[Int]
+
+    println(query)
```
khemrajrathore, line 37: remove println

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
```diff
@@ -0,0 +1,100 @@
...
+              property.sourceFileOut.head,
+              parsedQueryItem,
+              query,
+              property.lineNumber.getOrElse(-1).asInstanceOf[Int],
+              queryOrder
+            )
+          }
+        case None =>
+          println("failed")
+          logger.debug("Failed to parse")
```
khemrajrathore, line 54: remove println and have some description message in debug to understand what has failed

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
```diff
@@ -0,0 +1,100 @@
...
+          None
+      }
+    } catch {
+      case ex: Exception =>
+        println(s"Error while parsing SQL query at line $lineNumber: ${ex.getMessage}")
```
khemrajrathore, line 59: remove println

src/main/scala/ai/privado/passes/PropertyParserPass.scala
karan-batavia: In this case, we'll have to change the structure for all the parsers because we get line numbers for only `yml` at the moment.

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
karan-batavia: Property Node has `Option[None]` for line and column by default. Do we still need to store it?

khemrajrathore (dismissed):
LGTM

src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -169,30 +175,37 @@ class PropertyParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache, la
...
-        } catch {
-          case e: Throwable =>
-            logger.trace(s"Error while creating properties node for file $file")
-            logger.debug(s"Error while creating properties node for file : $file, error : ${e.getMessage}")
-            List[("", "")]()
-        }
-      }
-      case Left(error) => {
-        List[("", "")]()
+
```
khemrajrathore, line 178: as linenumber is also returned along with the value part here, can't we change the output to (String, String, linenumber) instead ?

karan-batavia: Ah yes, makes sense.

src/main/scala/ai/privado/passes/SQLPropertyPass.scala
```diff
@@ -0,0 +1,98 @@
...
+              builder,
+              property.sourceFileOut.head,
+              parsedQueryItem,
+              query,
+              lineNumber,
+              queryOrder
+            )
+          }
+        case None =>
+          println("Failed to parse query")
```
khemrajrathore, line 51: remove the print statement

khemrajrathore (dismissed):
LGTM

pandurangpatil (changes requested):
Add unit tests as discussed yesterday

PR #575, opened 2023-05-29
add - support for running scala file at runtime

khemrajrathore:
The file External.scala will have 

```
new ExternalScript {
  override def process(..params) = ...logic
}
```

we can specify the external script path in SystemConfig in privado config


pandurangpatil (approved):
looks good to me

PR #576, opened 2023-05-30
JS support to parse dynamic script tags + .env, yaml config files

dbMundada:
* Add support to parse the dynamic script tags

* Add support for Yaml, yml & .env for JS


pandurangpatil (approved):
looks good

PR #608, opened 2023-06-21
updated dependency

hiteshbedre:


src/main/scala/ai/privado/languageEngine/python/processor/PythonProcessor.scala
```diff
@@ -51,6 +52,7 @@ object PythonProcessor {
           // Apply default overlays
           X2Cpg.applyDefaultOverlays(cpg)
           new ImportsPass(cpg).createAndApply()
+          new ImportResolverPass(cpg).createAndApply()
```
DavidBakerEffendi, line 55: This import resolver needs to be added for JavaScript too

DavidBakerEffendi: We need to add `io.joern.pysrc2cpg.DynamicTypeHintFullNamePass` before `PythonTypeRecoveryPass`

hiteshbedre: For JavaScript we are calling `io.joern.jssrc2cpg.JsSrc2Cpg#postProcessingPasses` method. Which has mentioned pass.

DavidBakerEffendi: @hiteshbedre ah yes I see

DavidBakerEffendi:
This may be causing Java ast crashing https://github.com/joernio/joern/commit/451d5519f58a01a37a3b7a21e11412789a74e34f

pandurangpatil (approved):
looks good

PR #626, opened 2023-07-04
Javascript fixes

dbMundada:
- Fix the JS property not getting populated in API flows
- Use pattern from Collection Express rule
- Remove FPs from import/require statements
- Add support for parsing block with method, path & handler for Collection


pandurangpatil (approved):
looks good

PR #628, opened 2023-07-06
Scala 3 Upgrade

DavidBakerEffendi:
Upgraded Joern and as minimal dependencies as possible to their compatible Scala 3 variant. Removed Mockito as it is unused and incompatible.


pandurangpatil (approved):
looks good.

PR #634, opened 2023-07-11
Release 11 July 2023: Scala 3 Upgrade

ankit-privado:
* Scala 3 Upgrade Upgraded Joern and as minimal dependencies as possible to their compatible Scala 3 variant.

Note: [Mockito has no Scala 3 compatibility](https://index.scala-lang.org/mockito/mockito-scala) (yet) so we will have to disable it or find an alternative. The bad news is that this project has no clear path to [Scala 3](https://github.com/mockito/mockito-scala/issues/364#issuecomment-1192749274)

* Removing Mockito

* Updated Scala format to Scala 3

* Removed Scala 2.13 spec from code quality workflow

* Fixed Scala 3 compilation issues

* removed snakeyaml dependency

---------

### DIFF
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -18,7 +18,7 @@ jobs:
           distribution: 'temurin'
           java-version: '17'
       - name: Check formatting
-        run: sbt ++2.13.8 scalafmtCheck test:scalafmtCheck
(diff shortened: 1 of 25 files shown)


pandurangpatil (approved):
Looks good

PR #641, opened 2023-07-13
Language engine/version update (#640)

karan-batavia:
* Updated privado language engine version to 2.0.13

* Removed unnecessary files

* remove unnecessary file

* change updateDependencies.sh to use scala 3

* Update underlying dependencies


pandurangpatil (approved):
looks good

PR #655, opened 2023-07-29
[rubysrc2cpg] Added input parameter to exclude source files

ankit-privado:
Added Support to exclude source file.


src/main/scala/ai/privado/entrypoint/CommandParser.scala
```diff
@@ -48,7 +48,8 @@ case class PrivadoInput(
   testOutput: Boolean = false,
   showUnresolvedFunctionsReport: Boolean = false,
   generateAuditReport: Boolean = false,
-  enableAuditSemanticsFilter: Boolean = false
+  enableAuditSemanticsFilter: Boolean = false,
+  excludeFileRegex: String = ""
```
khemrajrathore, line 52: @ankit-privado Instead of taking the file regex to exclude from command line, can we not use the currently defined exclusion rules in `privado` -> `config` -> 'exclusion'?
cc - @pandurangpatil

khemrajrathore: I would suggest to have a flag which is a boolean, which suggest whether we want to skip files present in exclusion rule or not

ankit-privado: yes @khemrajrathore right we can do this, but don't you think in this case we need to provide two flags and also need to handle all possible combinations? Also, in the future if we want to exclude both files and rules, then this approach won't work.

pandurangpatil: @ankit-privado What @khemrajrathore pointing at is makes sense. Let's use the exclusion rule to exclude the file. For now, we will handle it for Ruby, for other frontends, we will handle that in a separate change. Adding multiple configurations and input parameters adds up to lot of confusion later on. Let's reuse the existing configuration. 

@khemrajrathore I didn't get the purpose of the boolean flag. I guess we don't need boolean flag.

ankit-privado: okay okay got it. will make the change.

khemrajrathore: @pandurangpatil Sometimes we actually need to process entire cpg and tag only the actual code(excluding excluded files). There we may need this flag. We can better take input from the code review team. whether they need such a boolean flag

khemrajrathore: cc - @HiteshMah-Jan What is your thought on this? Can we use the exclusion regex directly or better to have a boolean flag

ankit-privado: Resolved.

PR #656, opened 2023-07-31
Ruby type recovery global symbol

khemrajrathore:


pandurangpatil (approved):
looks good

DavidBakerEffendi:
Interesting, I used to have a global symbol table in the earlier iterations of this but resorted to querying the CPG since the global symbol table would add additional context insensitivity which was difficult to resolve/manage. But this does confirm a thought I had that querying the CPG inter-procedurally takes large computational resources itself, and we may want to index some sub-graphs, e.g. members/types.

PR #672, opened 2023-08-09
DynamicTypeHintFullName for sink tagging

ankit-privado:
Added dynamicTypeHintName for sink tagging


src/main/scala/ai/privado/languageEngine/ruby/tagger/sink/RegularSinkTagger.scala
```diff
@@ -45,7 +45,10 @@ class RegularSinkTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelC
   }
 
   override def runOnPart(builder: DiffGraphBuilder, ruleInfo: RuleInfo): Unit = {
-    val sinks = cacheCall.methodFullName("(pkg.){0,1}(" + ruleInfo.combinedRulePattern + ").*").l
+    val combinedRegex = s"(pkg.){0,1}(\"${ruleInfo.combinedRulePattern}\").*"
```
khemrajrathore, line 48: We don't need `pkg.` in ruby, we can directly use `ruleInfo.combinedRulePattern`

ankit-privado: done.

PR #678, opened 2023-08-09
Ruby post processing pass args limit

pandurangpatil:
This PR wil require corresponding Language Engine changes to be released first.

### DIFF
--- a/src/main/scala/ai/privado/utility/Utilities.scala
+++ b/src/main/scala/ai/privado/utility/Utilities.scala
@@ -22,12 +22,16 @@
 
 package ai.privado.utility
 
-import ai.privado.cache.RuleCache
+import ai.privado.cache.{AppCache, RuleCache}
+import ai.privado.entrypoint.ScanProcessor
 import ai.privado.metric.MetricHandler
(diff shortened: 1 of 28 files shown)


src/main/scala/ai/privado/utility/Utilities.scala
```diff
@@ -62,6 +66,20 @@ object Utilities {
 
   private val logger = LoggerFactory.getLogger(getClass)
 
+  def getEngineContext(maxCallDepthP: Int = 4)(implicit semanticsP: Semantics = DefaultSemantics()): EngineContext = {
+    val expanLimit =
+      if ScanProcessor.config.limitArgExpansionDataflows > -1 then ScanProcessor.config.limitArgExpansionDataflows
```
khemrajrathore, line 71: Instead of using the `ScanProcessor.config` as a static variable, better to pass it in method `getEngineContext`, as it con create conflicting result in writing testcases, due to ScanProcessor.config holding a single value

pandurangpatil: done

PR #684, opened 2023-08-15
[RUBY] - Parallel download dependency

pandurangpatil:
1. Abstract ConcurrentProcessor with ZeorMQ queue to write the final results to a single result object in an independent thread.
2. New Download dependency pass implementation, with concurrent download and unpack implementation by extending ConcurrentProcessor.
3. Downloading the dependencies of the dependencies in a recursive loop.
4. Processing downloaded ruby files to generate PackageTable by using
new ConcurrentProcessor with ZeroMQ queue.

TODO
1. Add unit tests at method level.
2. PackageTable is set inside Singleton object. Which will result into
holding the respective memory block. Either remove it from Singleton
object or provide a cleanup method to empty the table post the work is
done.


khemrajrathore:
Merging this branch to `addSchemaViolation`

PR #687, opened 2023-08-16
Ruby august release

khemrajrathore:
- Language engine upgrade to version `2.0.52`
- Ruby types for external packages
- Cfg exception handling for ruby cpg creation
- Properties Pass
- Derived sources from `schema.rb` file
- Parallelized Downloaded Dependency in ruby
- Considering dynamicTypeHintFullName while tagging sinks


pandurangpatil (approved):
looks good

PR #700, opened 2023-08-27
Add this FP fix for JS lang

dbMundada:


pandurangpatil (approved):
Missing sink from the comparison report is expected as per @dbMundada analysis. Going ahead with the merge

PR #709, opened 2023-09-01
patch - remove the dataflow's reachableBy implementation from Read pass

khemrajrathore:
There is a piece of code in `databaseReadPass` which runs Dataflow in runOnPart which is a resource-intensive task and becomes a bottleneck for repos like `openmrs-core` , I currently have added a flag  `disable-read-dataflow` and am able to get the repo scanned on the laptop using that flag. 
 
 Default behaviour is to run the read dataflow, but for repos which need a higher configuration `disable-read-dataflow` can be used to skip the intensive task
 
 The optimized version will be available in a follow-up PR


pandurangpatil (approved):
looks good to me

PR #712, opened 2023-09-05
Update Taggercache for Concurrent executions

dbMundada:


src/main/scala/ai/privado/cache/TaggerCache.scala
```diff
@@ -24,32 +24,46 @@
...
-
+import scala.collection.concurrent.TrieMap
 import scala.collection.mutable
-import java.util.concurrent.ConcurrentHashMap
-import scala.jdk.CollectionConverters.ConcurrentMapHasAsScala
 
 class TaggerCache {
   // Stores typeDeclFullName --> ( sourceRuleId --->  Member Node)
-  val typeDeclMemberCache = new ConcurrentHashMap[String, mutable.HashMap[String, mutable.HashSet[Member]]]().asScala
+  val typeDeclMemberCache = TrieMap[String, TrieMap[String, mutable.Set[Member]]]()
```
khemrajrathore, line 32: @dbMundada Can we also have a corresponding concurrent DS for `mutable.Set` ?

src/main/scala/ai/privado/cache/TaggerCache.scala
```diff
@@ -24,32 +24,46 @@
...
-      typeDeclMemberCache(typeDeclVal).addOne(ruleId -> mutable.HashSet())
-    typeDeclMemberCache(typeDeclVal)(ruleId).addOne(typeDeclMember)
+    typeDeclMemberCache.getOrElseUpdate(typeDeclVal, TrieMap[String, mutable.Set[Member]]()).updateWith(ruleId) {
+      case Some(existingMembers) => Some(existingMembers += typeDeclMember)
+      case None                  => Some(mutable.Set(typeDeclMember))
+    }
+  }
+
+  def getTypeDeclMemberCacheItem(typeDeclVal: String): Map[String, Set[Member]] = {
+    typeDeclMemberCache.getOrElse(typeDeclVal, TrieMap.empty).view.mapValues(_.toSet).toMap
```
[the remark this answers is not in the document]
dbMundada: Yes, its needed. Without that .mapValues is going to be deprecated.

src/main/scala/ai/privado/cache/TaggerCache.scala
dbMundada: Yes, I also had same thought but got this:

 **scala.collection.mutable.Set[Member]** is used within the **TrieMap**, and access to it is synchronized. This ensures that the inner sets are thread-safe for concurrent updates within the TrieMap, addressing potential concurrency issues when multiple threads modify the same set concurrently.

khemrajrathore (approved):
LGTM

PR #716, opened 2023-09-06
Ruby API Flows

khemrajrathore:
- Removed NaiveCalLinker for Ruby
- Upgraded to language engine version 2.0.78
- Increased Dataflow parameter expansion constant from 20 to 100
- Refactor in API Tagger of Ruby


src/main/scala/ai/privado/model/Constants.scala
```diff
@@ -162,5 +162,5 @@ object Constants {
   val semanticDelimeter = "_A_"
 
   // Ruby defaults
-  val defaultExpansionLimit = 20
+  val defaultExpansionLimit = 100
```
dbMundada, line 165: What is impact of this change?

src/main/scala/ai/privado/languageEngine/ruby/tagger/source/IdentifierTagger.scala
```diff
@@ -27,12 +27,19 @@ import ai.privado.cache.RuleCache
...
+import io.shiftleft.codepropertygraph.generated.{Cpg, Operators}
 import io.shiftleft.passes.ForkJoinParallelCpgPass
-import io.shiftleft.semanticcpg.language._
+import io.shiftleft.semanticcpg.language.*
+
+import scala.jdk.CollectionConverters.CollectionHasAsScala
 
 class IdentifierTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelCpgPass[RuleInfo](cpg) {
 
+  val cachedCall: List[Call] = cpg.call
```
dbMundada, line 39: Where is this used?

src/main/scala/ai/privado/model/Constants.scala
khemrajrathore: Increase in this limit allows us to generate more tasks in dataflow operation, In dataflow engine the default limit is 1000

src/main/scala/ai/privado/languageEngine/ruby/tagger/source/IdentifierTagger.scala
khemrajrathore: Oh, this is not used anywhere. Will need to remove this. 
Will remove this in subsequent PR's as a new commit will again take time for comparison report

pandurangpatil (approved):
look good to me.

PR #725, opened 2023-09-20
fix - rules getting added twice

khemrajrathore:
The PR fixes the following
- Rules getting processed twice
- Too many property nodes getting generated via propertyPass


dbMundada (approved):
Lgtm

PR #739, opened 2023-09-25
go: added derived source support

hiteshbedre:


ankit-privado:
Looking good to me, are you planning to add unit tests?

hiteshbedre:
> Looking good to me, are you planning to add unit tests?

Test case will be part of upcoming PR.

ankit-privado (approved):

build.sbt
```diff
@@ -8,7 +8,7 @@ ThisBuild / version      := sys.env.getOrElse("BUILD_VERSION", "dev-SNAPSHOT")
 
 
 val cpgVersion        = "1.4.22"
-val joernVersion      = "2.0.92"
```
khemrajrathore, line 11: Don't upgrade to 98 we are seeing delta in comparison report

khemrajrathore (commented):
@hiteshbedre Include test case as well, otherwise it is difficult to understand whether the sources we are looking at will be getting tagged or not

hiteshbedre:
> @hiteshbedre Include test case as well, otherwise it is difficult to understand whether the sources we are looking at will be getting tagged or not

Added test cases here: d0220250af68fffb52af29ccbfdce5de190e6f0d
Covered cases:
1. Member Node tagging
2. Object having tagged member tagging.

build.sbt
hiteshbedre: do you know which joern version is stable at current state? anyways we need to update to latest as Go's fixes are on the way.

khemrajrathore: we can upgrade till 2.0.92, I guess there is already a PR created by Datta on Dev

pandurangpatil:
@khemrajrathore let the Joern upgrade done for this branch of `go`. We need to anyway get the diff issue resolved before the GO release before 6th Oct.

cc: @hiteshbedre

khemrajrathore (approved):
LGTM

PR #742, opened 2023-09-26
Update the HTMLParser to adopt identifying script tags

dbMundada:
- Update HTMLParser to get HTML Elements excluding the script tag errors
- Update JSAPITagger to adopt HTMLElement in parsing
- Convert Inline script tag code into `content` attribute

### DIFF
--- a/src/main/scala/ai/privado/passes/HTMLParserPass.scala
+++ b/src/main/scala/ai/privado/passes/HTMLParserPass.scala
@@ -28,6 +28,7 @@ import ai.privado.model.Constants
 import ai.privado.tagger.PrivadoParallelCpgPass
 import ai.privado.utility.Utilities
 import com.gargoylesoftware.htmlunit.html._
+import com.gargoylesoftware.htmlunit.html.HtmlScript
 import com.gargoylesoftware.htmlunit.{BrowserVersion, WebClient}
 import io.shiftleft.codepropertygraph.generated.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.{NewFile, NewTemplateDom}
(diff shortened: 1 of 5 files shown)


src/main/scala/ai/privado/passes/HTMLParserPass.scala
```diff
@@ -85,6 +86,12 @@ class HTMLParserPass(cpg: Cpg, projectRoot: String, ruleCache: RuleCache) extend
       val webClient = new WebClient(
         new BrowserVersion.BrowserVersionBuilder(BrowserVersion.CHROME).setOnLine(false).build()
       )
+      val options = webClient.getOptions
+
+      // Configure options to ignore certain errors
+      options.setThrowExceptionOnFailingStatusCode(false) // Do not throw exceptions for failing HTTP status codes
+      options.setThrowExceptionOnScriptError(false)       // Do not throw exceptions for JavaScript errors
```
khemrajrathore, line 93: Does this setting help in minimizing the HTML related errors we see when parsing code?

dbMundada: Without this if any HTML file we getting errors we loosing all the HTML Elements inside that file.
But with this setting even though the errors come but we will able to get those HTML Elements inside it.

Not going to remove any errors we are showing in console.

dbMundada: https://github.com/Privado-Inc/privado-core/pull/742/files#diff-b5040cb043e63e7c73b2468b4d9c3891539c8b4a22f18dc66166c58920b2348eL93

We can remove console logs if we change this one.

khemrajrathore: log.debug is fine

khemrajrathore:
@pandurangpatil Looks good to me, please have a quick eye over this

pandurangpatil (approved):
looks good to me

PR #744, opened 2023-09-26
increase zeromq sockets

tuxology:
Fixes download dependency issue we sometimes face with large dependency list
- [x] Test on large repo

### DIFF
--- a/src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
+++ b/src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
@@ -13,7 +13,7 @@ import scala.concurrent.{Await, Future}
 object ConcurrentProcessor {
   val zContext = new ZContext()
   zContext.getContext.setMaxSockets(4096)
-  val STOP     = "STOP"
+  val STOP = "STOP"
 }
 
 /** Concurrent processor to process the individual processing of the data in parallel threads for each item of the array
--- a/src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
+++ b/src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
@@ -49,8 +49,8 @@ abstract class ConcurrentProcessor[T, R](result: R, rules: RuleCache) {
 
   // set maximum sockets that can be used
   private val count          = rules.getSystemConfigByKey(Constants.MAX_SOCKET_COUNT, true)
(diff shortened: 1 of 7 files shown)


src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
```diff
@@ -12,7 +12,8 @@ import scala.concurrent.{Await, Future}
 
 object ConcurrentProcessor {
   val zContext = new ZContext()
-  val STOP     = "STOP"
+  zContext.getContext.setMaxSockets(4096)
```
pandurangpatil, line 15: @tuxology I guess we should take this value as an input to CLI by setting this value of 4096 as a default value to that input.

tuxology: Updated and tested this with YAML on pay repo. We do need to specific the value in string for now, eg "2048". Default has been set for 4096 and upper limit is 10,000

src/main/scala/ai/privado/utility/ConcurrentProcessor.scala
```diff
@@ -39,13 +40,19 @@ object ConcurrentProcessor {
...
   private val logger = LoggerFactory.getLogger(this.getClass)
   // PUSH socket maintained for each task thread (Future)
   private val queue: ThreadLocal[ZMQ.Socket] = new ThreadLocal[ZMQ.Socket]
   // Unique ZeroMQ inproc:// topic used for each instance of ConcurrentProcessor.
   private val topic: String = UUID.randomUUID().toString
 
+  // set maximum sockets that can be used
+  private val count          = rules.getSystemConfigByKey(Constants.MAX_SOCKET_COUNT)
+  private val maxSocketCount = if (count.isEmpty) 1024 else if (count.toInt > 4000) 1024 else count.toInt
+  println(maxSocketCount)
```
khemrajrathore: @tuxology Can we have a little descriptive message along with socket count to understand this on console?

khemrajrathore: Nevermind was refering a older commit

khemrajrathore:
@tuxology We will need a corresponding PR on privado with the default value being set there in systemConfig

pandurangpatil (approved):
looks good

PR #745, opened 2023-09-28
[Fix] Dedup on intersecting path ids

karan-batavia:
We were seeing identical `pathIds` in the dataflow section of the violations. Used `dedup` on the `pathIds` to include each pathId only once.


khemrajrathore:
@karan-batavia Policy test case is failing

src/main/scala/ai/privado/cache/DataFlowCache.scala
```diff
@@ -138,7 +138,8 @@ object DataFlowCache {
     }
 
     if (ScanProcessor.config.generateAuditReport) {
-      AuditCache.addIntoBeforeFirstDedup(dataflow)
+      println("Generate")
```
khemrajrathore, line 141: remove the print statement

khemrajrathore (approved):
LGTM

PR #746, opened 2023-09-29
Dev

khemrajrathore:
- Update the HTMLParser to adopt identifying script tags
- Fix for dedup on intersecting path Ids
- Update the threats to export violations with unique sourceId
- Add offline flag to optionally download JS files in HTML Pass
- Update code engine version to 2.0.105


dbMundada (approved):
LGTM

PR #773, opened 2023-10-11
Fixes for string; domain + JS identifier api support in script tags

dbMundada:

PR #776, opened 2023-10-11
Dev

dbMundada:
…#773)

* Fixes for string; domain + JS identifier api support in script tags

* Add test-cases for script tag tagging

---------


pandurangpatil (approved):
looks good

PR #579, opened 2023-05-31
Support additional threats & Policies

dbMundada:

PR #781, opened 2023-10-12
Fix - Bug in export

khemrajrathore:
On the customer repo, we observed a weird bug while exporting, while debugging found it was introduced due to the PR - https://github.com/Privado-Inc/privado-core/pull/579

Have done some refactoring so that the code follows more stands of scala

cc - @pandurangpatil

### DIFF
--- a/src/main/scala/ai/privado/exporter/PolicyAndThreatExporter.scala
+++ b/src/main/scala/ai/privado/exporter/PolicyAndThreatExporter.scala
@@ -74,15 +74,8 @@ class PolicyAndThreatExporter(
     )
   }
 
-  private def convertProcessingSources(sourceNode: (String, CfgNode)) = {
-    try {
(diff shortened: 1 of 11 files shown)


src/main/scala/ai/privado/exporter/PolicyAndThreatExporter.scala
```diff
@@ -74,15 +74,8 @@ class PolicyAndThreatExporter(
     )
   }
 
-  private def convertProcessingSources(sourceNode: (String, CfgNode)) = {
-    try {
-      ViolationProcessingModel(sourceNode._1, ExporterUtility.convertIndividualPathElement(sourceNode._2).get, None)
-    } catch {
-      case e: Exception =>
-        logger.debug("Exception : ", e)
-        ViolationProcessingModel(sourceNode._1, null, None)
```
khemrajrathore, line 83: Usage of `null` was the problem to be precise

pandurangpatil (approved):
looks good

PR #783, opened 2023-10-13
Fix - Bug in export (#781)

khemrajrathore:
* fix - error while exporting on some repos

* fix and refactoring


pandurangpatil (approved):
looks good

PR #786, opened 2023-10-25
Fix: External script processing to work with scala 3

karan-batavia:
Using dotty `ScriptEngine` instead of scala reflection to run the external script. 
@khemrajrathore , please review.


khemrajrathore:
- [ ] Add test case which verifies the behaviour of reading external script, so in future we can catch this effectively

Question - If script execution fails does scan continue or fails?

karan-batavia:
> Question - If script execution fails does scan continue or fails?

@khemrajrathore, the regular scan will continue.

karan-batavia:
> * [x]  Add test case which verifies the behaviour of reading external script, so in future we can catch this effectively
This is done, please review @khemrajrathore.

khemrajrathore (dismissed):
LGTM

khemrajrathore (approved):
LGTM

PR #788, opened 2023-10-27
Re-introduced Cross-Closure DDG Edges via Pass

DavidBakerEffendi:
Adding `ExperimentalLambdaDataFlowSupportPass` to add the `REACHING_DEF` edges back in that were reverted in Joern.


khemrajrathore:
@DavidBakerEffendi Let's also increase the versions to 
```
val cpgVersion        = "1.4.25"
val joernVersion      = "2.0.131"
```

so that we can make sure, we are not having any missing flows

PR #792, opened 2023-10-30
Update for ruby package change

tuxology:
Adapting our core to use the `deprecated` package for the existing ruby frontend based on recent changes upstream: https://github.com/joernio/joern/pull/3755 This was tested on a few ruby repos like `pay` and `olubalance`


pandurangpatil (approved):
Looks good to me

PR #795, opened 2023-11-01
Go: sql schema node for queries written in go files

hiteshbedre:


src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+        buildAndAddSqlQueryNodes(queryLiteral, builder, fileNode)
+  }
+
+  private def buildAndAddSqlQueryNodes(
+    queryLiteral: Literal,
+    builder: DiffGraphBuilder,
+    fileNode: NodeOrDetachedNode
+  ): Unit = Try {
+
+    val queryLineNumber = queryLiteral.lineNumber.getOrElse(Integer.valueOf(0))
```
khemrajrathore, line 81: If the lineNumber is missing better to use `-1
`

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+}
+import io.shiftleft.codepropertygraph.generated.{Cpg, EdgeTypes, Operators}
+import org.slf4j.LoggerFactory
+import overflowdb.{BatchedUpdate, NodeOrDetachedNode}
+import io.shiftleft.semanticcpg.language.*
+
+import scala.collection.mutable
+import scala.util.{Failure, Success, Try}
+
+class SQLQueryParser(cpg: Cpg) extends PrivadoParallelCpgPass[Literal](cpg) {
```
khemrajrathore, line 48: @hiteshbedre Is this a copied file, I see nearly same implementation already done for JS/Python, There I believe the only difference is we are not creating SQL nodes. We can reimplement it more like OOPS and avoid repeatation for future as well

src/main/scala/ai/privado/model/Constants.scala
```diff
@@ -167,4 +167,6 @@ object Constants {
 
   // Ruby defaults
   val defaultExpansionLimit = 100
+
+  val dummyFileName = "dummyFileName"
```
khemrajrathore, line 171: Instead of calling this `dummyFileName` better we call it `unknown`, so this constant can be used across for all these kind of cases

src/main/scala/ai/privado/languageEngine/go/processor/GoProcessor.scala
```diff
@@ -57,6 +58,8 @@ object GoProcessor {
             s"${TimeMetric.getNewTime()} - Run oss data flow is done in \t\t\t- ${TimeMetric.setNewTimeToLastAndGetTimeDiff()}"
           )
 
+          new SQLParser(cpg, sourceRepoLocation, ruleCache).createAndApply()
+          new SQLQueryParser(cpg).createAndApply()
```
khemrajrathore, line 62: Are we also calling SQLQueryTagger in Go's PrivadoTagger?

khemrajrathore: This will help in tagging SQL nodes which match sources regex

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  val sqlQueryRegexPattern = "(?i).*(SELECT|INSERT|UPDATE|DELETE|CREATE|ALTER|DROP|TRUNCATE).*"
+  val logger               = LoggerFactory.getLogger(getClass)
+  override def generateParts(): Array[_ <: AnyRef] = {
+    //    CPG query to fetch the Literal with SQL string
+    //    'Repeat until' is used to combine multiline SQL queries into one
+    cpg.literal
+      .code(sqlQueryRegexPattern)
+      .repeat(_.astParent)(_.until(_.isCall.whereNot(_.name(s"${Operators.addition}|<operator>.stringExpressionList"))))
```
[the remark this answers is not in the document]
hiteshbedre: I have tested scenarios such as plain string and concatenated string given below, please let me know if any case Im missing:

case 1: string concatenation
createVotes := "CREATE TABLE IF NOT EXISTS votes (" +
        "id SERIAL NOT NULL," +
        "created_at datetime NOT NULL," +
        "candidate VARCHAR(6) NOT NULL," +
        "PRIMARY KEY (id)" +
        ");"


case 2: plain string
createVotes := `CREATE TABLE IF NOT EXISTS votes ( 
		id SERIAL NOT NULL, 
		created_at datetime NOT NULL, 
		candidate VARCHAR(6) NOT NULL, 
		PRIMARY KEY (id) 
	);`

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
hiteshbedre: Addressed here: f71787727ca546e68a532a05bdd5ae8a887c7e43

src/main/scala/ai/privado/model/Constants.scala
hiteshbedre: Agree, addressed here: c043eecc1ef51555209fc287a5bc7bc408f969ec

src/main/scala/ai/privado/languageEngine/go/processor/GoProcessor.scala
hiteshbedre: It wasn't added previously, addressed(added) in commit: ab05f8ed6487167bcdae4e683e8f4be680fd4f50

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+        buildAndAddSqlQueryNodes(queryNode, builder, fileNode)
+  }
+
+  private def buildAndAddSqlQueryNodes(
+    queryLiteral: AstNode,
+    builder: DiffGraphBuilder,
+    fileNode: NodeOrDetachedNode
+  ): Unit = Try {
+
+    val queryLineNumber = queryLiteral.lineNumber.getOrElse(Integer.valueOf(-1))
```
pandurangpatil, line 81: Ok so we are using Literal line number. While handling the same for ORM classes, we need to make sure to add line numbers of member variable declaration along with file node.

hiteshbedre:
@khemrajrathore any update on this PR?

khemrajrathore (approved):
LGTM

PR #801, opened 2023-11-02
Upgrade core engine version to 140

dbMundada:


src/main/scala/ai/privado/languageEngine/ruby/processor/RubyProcessor.scala
```diff
@@ -342,7 +342,7 @@ object RubyProcessor {
     val xtocpg = withNewEmptyCpg(cpgconfig.outputPath, cpgconfig: Config) { (cpg, config) =>
 
       new MetaDataPass(cpg, Languages.RUBYSRC, config.inputPath).createAndApply()
-      new ConfigFileCreationPass(cpg).createAndApply()
```
pandurangpatil, line 345: Whats this change about?

PR #805, opened 2023-11-02
go: sql nodes for gorm framework

hiteshbedre:
Description

We are building graph for SQL nodes. In Gorm framework a module defined is act as table in sql and its member as column. So we are capturing TypeDecl nodes on whom gorm related functions are written.
Ex.

      import (
		"github.com/jinzhu/gorm"
	)

	type User struct {
		// we need to use our own custom model because the id is different of the "gorm.Model"
		ID      string `gorm:"PRIMARY KEY; UNIQUE" json:"id"`
		Name    string `gorm:"type:varchar(255); NOT NULL" json:"name" validate:"required"`
		Email   string `gorm:"type:varchar(255)" json:"email"`
		Phone   string `gorm:"type:varchar(100); NOT NULL; UNIQUE; UNIQUE_INDEX" json:"phone" validate:"required"`
		Address string `gorm:"type:text" json:"address"`
		MyModel
	}

	// Save saves a User inside the database
	func (u *User) Save(db *gorm.DB) error {
		uuidResult, err := uuid.NewRandom()
		if err != nil {
			log.Fatal(err)
		}

		u.ID = uuidResult.String()
		u.CreatedAt = time.Now().String()
		u.UpdatedAt = time.Now().String()

		err = db.Save(&u).Error
		if err != nil {
			return err
		}
		return nil

In above code snippet we are capturing method node having parameter `gorm.DB` and processing its corresponding `TypeDecl` node to create sql related nodes.

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
+++ b/src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
@@ -1,0 +1,75 @@
+package ai.privado.languageEngine.go.passes.orm
+
+import ai.privado.cache.RuleCache
+import ai.privado.model.Constants
+import ai.privado.model.sql.{SQLColumn, SQLQuery, SQLQueryType, SQLTable}
+import ai.privado.tagger.PrivadoParallelCpgPass
+import ai.privado.utility.SQLNodeBuilder
+import ai.privado.utility.SQLParser.createSQLTableItem
+import better.files.*
+import io.joern.x2cpg.SourceFiles
+import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.codepropertygraph.generated.{Cpg, EdgeTypes}
+import io.shiftleft.semanticcpg.language.*
+import org.slf4j.LoggerFactory
+import overflowdb.{BatchedUpdate, NodeOrDetachedNode}
+
+import scala.collection.mutable
+import scala.util.{Failure, Success, Try}
+
+class GormParser(cpg: Cpg) extends PrivadoParallelCpgPass[TypeDecl](cpg) {
+  val GORM_PARAMETER_TYPE_RULE = ".*github.com/jinzhu/gorm.*"
+  val logger                   = LoggerFactory.getLogger(getClass)
+
+  override def generateParts(): Array[_ <: AnyRef] = {
+    cpg.method.where(_.parameter.typeFullName(GORM_PARAMETER_TYPE_RULE)).typeDecl.dedup.toArray
+  }
+
+  override def runOnPart(builder: DiffGraphBuilder, model: TypeDecl): Unit = {
+    Try(model.file.head) match {
+      case Success(fileNode) =>
+        buildAndAddSqlQueryNodes(builder, model, fileNode)
+      case Failure(_) =>
+        val fileNode = NewFile().name(Constants.Unknown)
+        buildAndAddSqlQueryNodes(builder, model, fileNode)
+    }
+  }
+
+  private def buildAndAddSqlQueryNodes(
+    builder: DiffGraphBuilder,
+    model: TypeDecl,
+    fileNode: NodeOrDetachedNode
+  ): Unit = {
+    Try {
+      try {
+        val sqlTable: SQLTable = SQLTable(
+          model.code.stripPrefix("\"").stripSuffix("\""),
+          model.lineNumber.getOrElse(Integer.valueOf(-1)),
+          model.columnNumber.getOrElse(Integer.valueOf(-1))
+        )
+        val sqlColumns: List[SQLColumn] = model.member.l.map(x =>
+          SQLColumn(
+            x.code.stripPrefix("\"").stripSuffix("\""),
+            x.lineNumber.getOrElse(Integer.valueOf(-1)),
+            x.columnNumber.getOrElse(Integer.valueOf(-1))
+          )
+        )
+        val queryModel = SQLQuery(SQLQueryType.CREATE, sqlTable, sqlColumns)
+        SQLNodeBuilder.buildAndReturnIndividualQueryNode(
+          builder,
+          fileNode,
+          queryModel,
+          model.code,
+          model.lineNumber.getOrElse(Integer.valueOf(-1)),
+          0
+        )
+
+      } catch {
+        case ex: Exception =>
+          ex.printStackTrace()
+          println(s"Error while building sql nodes for GORM framework: ${ex.getMessage}")
+          None
+      }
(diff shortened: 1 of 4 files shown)


src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import scala.collection.mutable
+import scala.util.{Failure, Success, Try}
+
+class GormParser(cpg: Cpg) extends PrivadoParallelCpgPass[TypeDecl](cpg) {
+  val GORM_PARAMETER_TYPE_RULE = ".*github.com/jinzhu/gorm.*"
+  val logger                   = LoggerFactory.getLogger(getClass)
+
+  override def generateParts(): Array[_ <: AnyRef] = {
+    cpg.method.where(_.parameter.typeFullName(GORM_PARAMETER_TYPE_RULE)).typeDecl.dedup.toArray
```
khemrajrathore, line 47: change query to `cpg.typeDecl.where(_.method.parameter...)` as there will be always less typeDecl node as compared to method nodes

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
```diff
@@ -0,0 +1,93 @@
...
+      try {
+        val sqlTable: SQLTable = SQLTable(
+          model.code,
+          model.lineNumber.getOrElse(Integer.valueOf(-1)),
+          model.columnNumber.getOrElse(Integer.valueOf(-1))
+        )
+        val sqlColumns: List[SQLColumn] = model.member.l.map(x =>
+          SQLColumn(x.code, x.lineNumber.getOrElse(Integer.valueOf(-1)), x.columnNumber.getOrElse(Integer.valueOf(-1)))
+        )
+        val queryModel = SQLQuery(SQLQueryType.CREATE, sqlTable, sqlColumns)
```
khemrajrathore, line 75: will this be always be create Query?

khemrajrathore: or as this is a typeDecl node this is as good as Create query

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
```diff
@@ -0,0 +1,93 @@
...
+          model.code,
+          model.lineNumber.getOrElse(Integer.valueOf(-1)),
+          0
+        )
+
+      } catch {
+        case ex: Exception =>
+          ex.printStackTrace()
+          println(s"Error while building sql nodes for GORM framework: ${ex.getMessage}")
+          None
```
khemrajrathore, line 89: As the function returns Unit, we don't need to specify `None`

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
```diff
@@ -0,0 +1,106 @@
...
+    super.beforeAll()
+    X2Cpg.applyDefaultOverlays(cpg)
+
+    val context = new LayerCreatorContext(cpg)
+    val options = new OssDataFlowOptions()
+    new OssDataFlow(options).run(context)
+    new GormParser(cpg).createAndApply()
+  }
+
+  override val goFileContents =
```
khemrajrathore, line 45: In this test Case, which method gets matched with `cpg.method.where(_.parameter.typeFullName(GORM_PARAMETER_TYPE_RULE)).typeDecl.dedup.toArray` this query. ?
As I understand the Ty peDecl node we get will be of `User`

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
hiteshbedre: A model is reflection of table schema. In terms sql terminology create just suits better.

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
hiteshbedre: `Save` method will get matched to query. And `Save` is part of TypeDecl `User`.

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
hiteshbedre: Agree, addressed here: 245840ffc166800c843d782ae2efcc5abbb5d02f

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
hiteshbedre: Addressed here: 245840ffc166800c843d782ae2efcc5abbb5d02f

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
khemrajrathore: Ok, looks like this is handled as part of AST creation differently. 

My only concern is False positives due to this, I hope you have already tested this across different cases, if YES that should be good

hiteshbedre: Just to simplify purpose of query "Get all TypeDecl node which are dealing with GORM dependency".
TypeDecl node are only created when they actually defined in code base. As parse is tightly coupled with GORM I don't see any case where false positive may arise.

khemrajrathore (dismissed):
LGTM

src/main/scala/ai/privado/languageEngine/go/passes/orm/GormParser.scala
```diff
@@ -0,0 +1,92 @@
...
+        buildAndAddSqlQueryNodes(builder, model, fileNode)
+    }
+  }
+
+  private def buildAndAddSqlQueryNodes(
+    builder: DiffGraphBuilder,
+    model: TypeDecl,
+    fileNode: NodeOrDetachedNode
+  ): Unit = {
+    Try {
```
pandurangpatil, line 65: Why there are two consicutive `Try` blocks ?

hiteshbedre: First `Try` was to return Success/Failure object but we are not processing return object so removed `Try` block. `try` block is persisted.

khemrajrathore:
@hiteshbedre Why this PR was closed?

PR #814, opened 2023-11-03
Go: Threat: Threat evaluation on SQL Nodes

hiteshbedre:
Migrate few threat evaluators using EntityMapper to use SQL schema to make it more generic. Currently we have kept both EntityMapper and Sql Node evaluation logic inplace, but slowly we will remove it completely.


src/main/scala/ai/privado/utility/SQLParser.scala
```diff
@@ -210,7 +212,7 @@ object SQLNodeBuilder {
       val columnNode = NewSqlColumnNode()
         .name(queryColumn.name)
         .code(queryColumn.name)
-        .lineNumber(queryLineNumber + queryColumn.lineNumber - 1)
+        .lineNumber(queryColumn.lineNumber)
```
hiteshbedre, line 215: @khemrajrathore why we were adding `queryLineNumber` to column's linenumber?

khemrajrathore: To get the effective line number of the columnNode, in the file

khemrajrathore: queryColumn.linenumber will give the linenumber of column node w.r.t to the passed query, But we need the line number w.r.t file

khemrajrathore: @hiteshbedre

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+              queryOrder
+            )
+          }
+        case None =>
+          logger.debug("Failed to parse: ", query, " : ", queryLineNumber)
+          None
+      }
+    } catch {
+      case ex: Exception =>
+        println(s"Error while parsing SQL query at line $queryLineNumber: ${ex.getMessage}")
```
khemrajrathore, line 102: Change this println to logger.debug

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -0,0 +1,87 @@
...
+  override def generateParts(): Array[_ <: AnyRef] = ???
+  val logger                                       = LoggerFactory.getLogger(getClass)
+
+  override def runOnPart(builder: DiffGraphBuilder, model: TypeDecl): Unit = {
+    Try(model.file.head) match {
+      case Success(fileNode) =>
+        buildAndAddSqlQueryNodes(builder, model, fileNode)
+      case Failure(_) =>
+        val fileNode = NewFile().name(Constants.Unknown)
+        buildAndAddSqlQueryNodes(builder, model, fileNode)
```
khemrajrathore, line 50: I see this piece of code of handling and adding SQL node getting repeat. Can we have a function to take of this?

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -0,0 +1,87 @@
...
+  ): Unit = {
+    try {
+      val sqlTable: SQLTable = SQLTable(
+        model.name,
+        model.lineNumber.getOrElse(Integer.valueOf(-1)),
+        model.columnNumber.getOrElse(Integer.valueOf(-1))
+      )
+      val sqlColumns: List[SQLColumn] = model.member.l.map(member =>
+        SQLColumn(
+          member.code,
```
khemrajrathore, line 67: For member, I guess we should use `name` instead of code. As for example if member defined is `private String userName` the name will be `userName` but code will not be `userName`

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -0,0 +1,87 @@
...
+        fileNode,
+        queryModel,
+        model.code,
+        model.lineNumber.getOrElse(Integer.valueOf(-1)),
+        0
+      )
+    } catch {
+      case ex: Exception =>
+        ex.printStackTrace()
+        logger.error(s"Error while building SQL nodes: ${ex.getMessage}")
```
khemrajrathore, line 84: Also add a corresponding debug message with more details, like file or TypeDecl causing the problem. So that it is easier to debug

src/main/scala/ai/privado/languageEngine/go/passes/orm/GorpParser.scala
```diff
@@ -0,0 +1,47 @@
...
+package ai.privado.languageEngine.go.passes.orm
+
+import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.codepropertygraph.generated.{Cpg, EdgeTypes}
+import io.shiftleft.semanticcpg.language.*
+import org.slf4j.LoggerFactory
+
+class GorpParser(cpg: Cpg) extends BaseORMParser(cpg) {
+  val GORP_PARAMETER_TYPE_RULE = "github.com/go-gorp/gorp.*"
+  val ADDRESS_OF_OBJECT_SYMBOL = "*[]"
```
khemrajrathore, line 32: is this `*[]` or coding error?

src/main/scala/ai/privado/semantic/Language.scala
```diff
@@ -56,17 +56,44 @@ object Language {
...
+    def sqlColumn: Traversal[SqlColumnNode] = trav.out(EdgeTypes.AST).cast[SqlColumnNode]
   }
 
   implicit class StepsForPropertyForSqlColumnNode(val trav: Traversal[SqlColumnNode]) extends AnyVal {
     def file: Traversal[File] = trav.out(EdgeTypes.SOURCE_FILE).cast[File]
 
   }
+
+  implicit class SqlColumnObject(val node: SqlColumnNode) {
+    def sqlTable: Option[SqlTableNode] = {
```
khemrajrathore, line 74: Can we simply not write `node.start.in(EdgeTypes.AST).cast[SqlTableNode]` ?

src/main/scala/ai/privado/semantic/Language.scala
```diff
@@ -56,17 +56,44 @@ object Language {
...
+        if (prop.isInstanceOf[SqlTableNode]) {
+          return Some(prop.asInstanceOf[SqlTableNode])
+        }
+      }
+      None
+    }
+  }
+
+  implicit class SqlTableObject(val node: SqlTableNode) {
+    def sqlQuery: Option[SqlQueryNode] = {
```
khemrajrathore, line 87: Similarly for below code

src/main/scala/ai/privado/threatEngine/CookieConsentMgmtModule.scala
```diff
@@ -1,8 +1,8 @@
-package ai.privado.languageEngine.java.threatEngine
+package ai.privado.threatEngine
 
 import ai.privado.cache.{AppCache, DataFlowCache, RuleCache}
+import ThreatUtility.hasDataElements
```
khemrajrathore, line 4: Can we import this from the base package `ai.privado.threatEngine.ThreatUtility.hasDataElements` ?

src/main/scala/ai/privado/threatEngine/PIIHavingDifferentRetentionPeriod.scala
```diff
@@ -1,9 +1,9 @@
-package ai.privado.languageEngine.java.threatEngine
+package ai.privado.threatEngine
 
 import ai.privado.cache.TaggerCache
 import ai.privado.exporter.ExporterUtility
 import ai.privado.languageEngine.java.passes.read.EntityMapper
-import ai.privado.languageEngine.java.threatEngine.ThreatUtility.{getPIINameFromSourceId, hasDataElements}
+import ThreatUtility.{getPIINameFromSourceId, hasDataElements}
```
khemrajrathore, line 6: import from base

src/main/scala/ai/privado/threatEngine/PIIShouldNotBePresentInMultipleTablesWithSQL.scala
```diff
@@ -0,0 +1,88 @@
...
+    threat: PolicyOrThreat,
+    cpg: Cpg,
+    taggerCache: TaggerCache
+  ): Try[(Boolean, List[ViolationProcessingModel])] = Try {
+    val violatingFlows = ListBuffer[ViolationProcessingModel]()
+
+    if (hasDataElements(cpg)) {
+      val taggedSources = getSources(cpg)
+      taggedSources.foreach(groupedSource => {
+        val tableNames           = groupedSource._2.collect(col => col.sqlTable.get.name)
```
khemrajrathore, line 38: As SQLTable here is option, better to not use `get` and have proper handling

src/main/scala/ai/privado/threatEngine/ThreatEngineExecutor.scala
```diff
@@ -157,13 +157,24 @@ class ThreatEngineExecutor(
...
         }
+        val result2 = PIIShouldNotBePresentInMultipleTables.getViolations(threat, cpg, taggerCache) match {
+          case Success(res) => Some(res)
+          case Failure(e) => {
+            logger.debug(s"Error for ${threatId}: ${e}")
+            None
+          }
+        }
+        (result1, result2) match {
+          case (Some((b1, list1)), Some((b2, list2))) => Some((b1 || b2, list1 ++ list2))
```
khemrajrathore, line 175: As I understand the 2 are basically present because we need to migrate. So both will not result in Some all the time. If 1 return Some, then 2nd will return None. Add those cases as well

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
```diff
@@ -0,0 +1,107 @@
...
+
+class GormParserTest extends GoTaggingTestBase {
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    X2Cpg.applyDefaultOverlays(cpg)
+
+    val context = new LayerCreatorContext(cpg)
+    val options = new OssDataFlowOptions()
+    new OssDataFlow(options).run(context)
```
khemrajrathore, line 41: running dataflow is not needed for this test Case right?

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
```diff
@@ -0,0 +1,107 @@
...
+
+import ai.privado.languageEngine.go.tagger.GoTaggingTestBase
+import ai.privado.model.*
+import ai.privado.semantic.Language.*
+import io.joern.x2cpg.X2Cpg
+import io.shiftleft.semanticcpg.language.*
+import io.shiftleft.semanticcpg.layers.LayerCreatorContext
+import io.joern.dataflowengineoss.layers.dataflows.{OssDataFlow, OssDataFlowOptions}
+
+class GormParserTest extends GoTaggingTestBase {
```
khemrajrathore, line 33: GoTaggingTestBase and be modified to have `withOSSDataflow` which takes boolean, and pass can be applied accordingly

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
```diff
@@ -0,0 +1,112 @@
...
+
+class GorpParserTest extends GoTaggingTestBase {
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    X2Cpg.applyDefaultOverlays(cpg)
+
+    val context = new LayerCreatorContext(cpg)
+    val options = new OssDataFlowOptions()
+    new OssDataFlow(options).run(context)
```
khemrajrathore, line 19: Similar comment as above

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
```diff
@@ -0,0 +1,112 @@
...
+      |	for _, u := range users {
+      |		fmt.Printf("ID: %d, Name: %s, Age: %d\n", u.ID, u.Name, u.Age)
+      |	}
+      |}
+      |
+      |""".stripMargin
+
+  "Adding sql nodes for GORP framework" should {
+    "check table nodes" in {
+      val tableNodes = cpg.sqlTable.l
```
khemrajrathore, line 87: Also add a test case for `sqlQuery` as well

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
```diff
@@ -0,0 +1,112 @@
...
+      tableNodes.head.name shouldBe "User"
+    }
+
+    "check column nodes" in {
+      val columnNodes = cpg.sqlColumn.l
+      columnNodes.size shouldBe 3
+
+      val List(id, name, age) = cpg.sqlColumn.l
+      id.code shouldBe "ID"
+      id.lineNumber.get shouldBe 15
```
khemrajrathore, line 98: This test case should ideally be `id.lineNumber shouldBe Some(15)`

src/test/scala/ai/privado/passes/SQLParserTest.scala
```diff
@@ -56,18 +60,64 @@ class SQLParserTest extends AnyWordSpec with Matchers with BeforeAndAfterAll {
...
+    "Query node check" in {
+      cpg.sqlQuery.size shouldBe 2
+      val List(a, b) = cpg.sqlQuery.l
+      a.name shouldBe SQLQueryType.CREATE
+      b.name shouldBe SQLQueryType.SELECT
+    }
+
+    "SqlTable node check" in {
+      val List(x) = cpg.sqlTable.dedupBy(_.name).l
+      x.name shouldBe "votes"
```
khemrajrathore, line 85: modify the test case to have 2 sqlTable nodes instead of deduping

src/test/scala/ai/privado/passes/SQLParserTest.scala
```diff
@@ -56,18 +60,64 @@ class SQLParserTest extends AnyWordSpec with Matchers with BeforeAndAfterAll {
...
+    }
+
+    "Traversal from query to table" in {
+      val List(x) = cpg.sqlQuery.sqlTable.dedupBy(_.name).l
+      x.name shouldBe "votes"
```
khemrajrathore, line 90: similar change as above

src/test/scala/ai/privado/passes/SQLParserTest.scala
```diff
@@ -56,18 +60,64 @@ class SQLParserTest extends AnyWordSpec with Matchers with BeforeAndAfterAll {
...
+    }
+
+    "Traversal from table to column" in {
+      val List(id, created_at, candidate) = cpg.sqlQuery.sqlTable.dedupBy(_.name).sqlColumn.l
```
khemrajrathore, line 94: Don't dedup, and verify that both tables have appropriate sqlColumn Nodes

src/test/scala/ai/privado/passes/SQLParserTest.scala
```diff
@@ -25,17 +25,21 @@ package ai.privado.passes
...
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.wordspec.AnyWordSpec
-import ai.privado.semantic.Language._
-import io.shiftleft.semanticcpg.language._
 
+import scala.collection.mutable
 class SQLParserTest extends AnyWordSpec with Matchers with BeforeAndAfterAll {
+  private val cpgs        = mutable.ArrayBuffer.empty[Cpg]
+  private val outPutFiles = mutable.ArrayBuffer.empty[File]
+  private val inputDirs   = mutable.ArrayBuffer.empty[File]
```
khemrajrathore, line 42: This can be part of a Base Test case as this will be needed for future test cases as well

src/test/scala/ai/privado/threatEngine/go/ThreatTests.scala
```diff
@@ -0,0 +1,237 @@
...
+      PolicyAction.DENY,
+      DataFlow(List(), SourceFilter(Option(true), ""), List()),
+      List("**"),
+      Map[String, String](),
+      Map[String, String](),
+      "",
+      Array[String]()
+    )
+
+    "When same data-element is part of multiple table in go file" in {
```
khemrajrathore, line 148: Is the description of this test case correct?

src/test/scala/ai/privado/threatEngine/go/ThreatTests.scala
```diff
@@ -0,0 +1,237 @@
...
+          |		return nil, err
+          |	}
+          |	return u, nil
+          |}
+          |
+          |""".stripMargin)
+
+      val result = threatEngine.processProcessingViolations(threat)
+      result should not be empty
+      result.get.policyId shouldBe "PrivadoPolicy.Storage.IsSamePIIShouldNotBePresentInMultipleTables"
```
khemrajrathore, line 125: Do we also need to pin point what are Data Elements were part of different tables? 
cc- @dbMundada

src/test/scala/ai/privado/threatEngine/sql/ThreatTests.scala
```diff
@@ -0,0 +1,113 @@
...
+    X2Cpg.applyDefaultOverlays(cpg)
+    val context = new LayerCreatorContext(cpg)
+    val options = new OssDataFlowOptions()
+    new OssDataFlow(options).run(context)
+    new SQLParser(cpg, config.inputPath, ruleCache).createAndApply()
+    new SqlQueryTagger(cpg, ruleCache).createAndApply()
+
+    new Dataflow(cpg).dataflow(privadoInput, ruleCache, dataFlowCache, auditCache)
+    val dataFlows: Map[String, Path] = Map()
+    new ThreatEngineExecutor(cpg, dataFlows, config.inputPath, ruleCache, null, dataFlowCache, privadoInput)
```
khemrajrathore, line 111: I believe all this can be part of a base Class, instead of repeating for every test class

src/test/scala/ai/privado/threatEngine/go/ThreatTests.scala
hiteshbedre: We have cases such as table definition exist in Go and SQL files. This test case belong to schema definition defined in .go file.

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private def buildAndAddSqlQueryNodes(
+    queryLiteral: AstNode,
+    builder: DiffGraphBuilder,
+    fileNode: NodeOrDetachedNode
+  ): Unit = Try {
+
+    val queryLineNumber = queryLiteral.lineNumber.getOrElse(Integer.valueOf(-1))
+    val query           = queryLiteral.code
```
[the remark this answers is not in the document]
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  private def buildAndAddSqlQueryNodes(
+    builder: DiffGraphBuilder,
+    model: TypeDecl,
+    fileNode: NodeOrDetachedNode
+  ): Unit = {
+    try {
+      val sqlTable: SQLTable = SQLTable(
+        model.name,
+        model.lineNumber.getOrElse(Integer.valueOf(-1)),
```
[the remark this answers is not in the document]
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
hiteshbedre: Agree
Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -0,0 +1,87 @@
...
+        builder,
+        fileNode,
+        queryModel,
+        model.code,
+        model.lineNumber.getOrElse(Integer.valueOf(-1)),
+        0
+      )
+    } catch {
+      case ex: Exception =>
+        ex.printStackTrace()
```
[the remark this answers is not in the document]
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c

src/main/scala/ai/privado/languageEngine/go/passes/orm/GorpParser.scala
hiteshbedre: not coding error. When object passed as pointer its signature come as `*[]packagename.StructName`

hiteshbedre: Added comment saying so in: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-acf64ac8318fa5936d0d13f010c3d12645fc5c510e04ae1428db5659f1ff1ee0

src/main/scala/ai/privado/threatEngine/CookieConsentMgmtModule.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-375232dc9bd151701124e2409d6be0e510254595781008fcc5979995326c9137R4

src/main/scala/ai/privado/threatEngine/PIIHavingDifferentRetentionPeriod.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-7667ab8534670cb21ce86d0d3760ff19a3a75ae25b93de7438faac9e4d591493R6

src/main/scala/ai/privado/threatEngine/PIIShouldNotBePresentInMultipleTablesWithSQL.scala
hiteshbedre: Handled better with `getOrElse`: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-e52d364abd76eb28a88e94a14440d5bc178f382802edceeec8130ad885fb9d6aR38

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
hiteshbedre: correct, removed from change: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-8945715e101c7f014dc6232057abf46c27728adba146a592a2e7e761702729f6L39

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
hiteshbedre: removed from change: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-e2d45390b76abf57ef0f0b153c177ff71cabdbc9f84c26bdbb4d9e16781e6b16L17

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
hiteshbedre: Added here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-e2d45390b76abf57ef0f0b153c177ff71cabdbc9f84c26bdbb4d9e16781e6b16R88

src/test/scala/ai/privado/languageEngine/go/passes/orm/GorpParserTest.scala
hiteshbedre: updated: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-e2d45390b76abf57ef0f0b153c177ff71cabdbc9f84c26bdbb4d9e16781e6b16R101

src/test/scala/ai/privado/passes/SQLParserTest.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-0b081aa0ef3414db6441e67f9aecd7f239cb721d546a0fa6d2a6e69f2529840bR84

src/test/scala/ai/privado/passes/SQLParserTest.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-0b081aa0ef3414db6441e67f9aecd7f239cb721d546a0fa6d2a6e69f2529840bR93

src/test/scala/ai/privado/threatEngine/go/ThreatTests.scala
hiteshbedre: Added data-element as part of test case: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-9b894724b8c80486fc9d176ab68ffe4e82cce2359fa4ff55b32c12abbda20d4dR126

src/test/scala/ai/privado/passes/SQLParserTest.scala
hiteshbedre: Addressed here: https://github.com/Privado-Inc/privado-core/pull/814/commits/1f896170d831e6bfac21ffbb33348b16be85211c#diff-0b081aa0ef3414db6441e67f9aecd7f239cb721d546a0fa6d2a6e69f2529840bR107

src/main/scala/ai/privado/utility/SQLParser.scala
hiteshbedre: Commit: 8c24bbdfc59ec27a7bd48753821c17f269946ef9 exclusively refactored lineNumber.
Logic:
1. While creating column node also passed Table node's object.
2. While initiating lineNumber if column added table's lineNumber to it.

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
hiteshbedre: Handled by inheritance.
Here is separate commit for same: c52d581324b2d67f0fa549e2b4938b90494ae209

src/main/scala/ai/privado/semantic/Language.scala
hiteshbedre: Addressed separately here: 203d820542cb72558c426009549788d534356d4a
@pandurangpatil please also review this commit.

src/main/scala/ai/privado/semantic/Language.scala
hiteshbedre: Addressed here: 203d820542cb72558c426009549788d534356d4a

src/main/scala/ai/privado/semantic/Language.scala
pandurangpatil: @khemrajrathore I guess the suggested code snippet is same as `node.in(EdgeTypes.AST).cast[SqlTableNode].head`. I didn't understood the purpose of `start` in this statement. 

In the event there is no parent node and if it returns an empty `Iterator` calling `.head` will throw error on it.

src/test/scala/ai/privado/languageEngine/go/passes/orm/GormParserTest.scala
hiteshbedre: Found dataflow isn't necessary for these test case. Removed it under: ea366f1f6499723e5f9c5835a13da604a66f9a17

src/test/scala/ai/privado/passes/SQLParserTest.scala
hiteshbedre: Moved to base class: ea366f1f6499723e5f9c5835a13da604a66f9a17

src/test/scala/ai/privado/threatEngine/sql/ThreatTests.scala
hiteshbedre: Moved all redundant test case related code in base class, here is single commit for same: ea366f1f6499723e5f9c5835a13da604a66f9a17

src/main/scala/ai/privado/languageEngine/go/passes/SQLQueryParser.scala
```diff
@@ -99,7 +92,7 @@ class SQLQueryParser(cpg: Cpg) extends PrivadoParallelCpgPass[AstNode](cpg) {
       }
     } catch {
       case ex: Exception =>
-        println(s"Error while parsing SQL query at line $queryLineNumber: ${ex.getMessage}")
+        logger.debug(s"Error while parsing SQL query at line $queryLineNumber: ${ex.getMessage}")
```
khemrajrathore, line 95: Also log the `query` and `fileName`, this will help in debugging

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -51,37 +52,37 @@ abstract class BaseORMParser(cpg: Cpg) extends PrivadoParallelCpgPass[TypeDecl](
     }
   }
 
-  private def buildAndAddSqlQueryNodes(
-    builder: DiffGraphBuilder,
-    model: TypeDecl,
-    fileNode: NodeOrDetachedNode
-  ): Unit = {
+  def buildAndAddSqlQueryNodes(builder: DiffGraphBuilder, model: AstNode, fileNode: NodeOrDetachedNode): Unit = {
+    val typeDeclNode = model.asInstanceOf[TypeDecl]
```
khemrajrathore, line 56: Taking AstNode as argument and typeCasting it to TypeDeclNode in base class isn't correct. Instead this function can be made abstract and the class extending this can use it as per there use case.
`def buildAndAddSqlQueryNodes(builder: DiffGraphBuilder, node: AstNode, fileNode: NodeOrDetachedNode): Unit = ???`

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
```diff
@@ -51,37 +52,37 @@ abstract class BaseORMParser(cpg: Cpg) extends PrivadoParallelCpgPass[TypeDecl](
...
         0
       )
     } catch {
       case ex: Exception =>
-        ex.printStackTrace()
         logger.error(s"Error while building SQL nodes: ${ex.getMessage}")
+        logger.debug(s"""Error while building SQL nodes for
+             |TypeDecl: name ${typeDeclNode.name} \n
+             |code: ${typeDeclNode.code} \n
+             |file: ${typeDeclNode.file.head.name}""".stripMargin)
```
khemrajrathore, line 85: use file.headOption.getOrElse

src/main/scala/ai/privado/semantic/Language.scala
khemrajrathore: @pandurangpatil Ideally we should avoid using `head` in this case as well and the return should be a traversal instead of a single node. 

Change the return type from `Option[SqlTableNode]` to `Traversal[SqlTableNode]` and similarly for other helper traversals as well

Cc - @hiteshbedre

src/main/scala/ai/privado/utility/SQLParser.scala
```diff
@@ -137,29 +133,29 @@ object SQLParser {
...
     val tableColumnNumber = Try(table.getASTNode.jjtGetFirstToken().beginColumn).getOrElse(NUMBER_MINUSONE)
     SQLTable(tableName, tableLineNumber, tableColumnNumber)
   }
 
-  private def createSQLColumnItem(column: ASTNodeAccess) = {
+  private def createSQLColumnItem(column: ASTNodeAccess, sqlTable: SQLTable) = {
     SQLColumn(
       column.toString,
-      Try(column.getASTNode.jjtGetFirstToken().beginLine).getOrElse(NUMBER_ONE),
+      Try(column.getASTNode.jjtGetFirstToken().beginLine).getOrElse(NUMBER_ONE) + sqlTable.lineNumber,
```
khemrajrathore, line 158: Is this yielding correct result, shouldn't this be `sqlTable.lineNumber - 1` ?

src/main/scala/ai/privado/semantic/Language.scala
pandurangpatil: @khemrajrathore it is expected to have only one parent to the given `SqlColumn` node.

what is the reasoning behind not calling `head` at this place as we know there is only a possibility of a single return node.

Either we do that at this place or we do it where we are going to consume this method. I would rather handle it at his place than doing it again and again everytime I consume this method.

khemrajrathore: @pandurangpatil Yes, the expectation is to get 1 parent, but in some cases if a sqlColumn node is created such that it has no parent of multiple parent (Due to some faulty implementation or bug) , that should also work smoothly and not throw runtime exception. 

And if you also refer to the core Language engine query they to refer to the same principle of using Traversal as much as possible. Not a blocker from my end its all about making queries robust

khemrajrathore: Also we want to do `head` instead of `Some(node.start.in(EdgeTypes.AST).cast[SqlTableNode].head)` we should do `node.start.in(EdgeTypes.AST).cast[SqlTableNode].headOption`

cc - @hiteshbedre

khemrajrathore: > @khemrajrathore I guess the suggested code snippet is same as `node.in(EdgeTypes.AST).cast[SqlTableNode].head`. I didn't understood the purpose of `start` in this statement.
> 
> In the event there is no parent node and if it returns an empty `Iterator` calling `.head` will throw error on it.

@pandurangpatil Won't this lead of `nullPointerException` if there is no parent ?

pandurangpatil: @khemrajrathore I agree with what you are pointing at when we are using the query like `cpg.sqlColumn.sqlTable.l` in that case it should be an iterator as you are pointing at. As that's going to be called a query.

In this case, the method is going to be called over an Object of `SqlColumnNode` and not on `Traversal[SqlColumnNode]`. If we do that over `Traversal[SqlColumnNode]` then certainly we need to handle it the way you are pointing.

pandurangpatil: > > @khemrajrathore I guess the suggested code snippet is same as `node.in(EdgeTypes.AST).cast[SqlTableNode].head`. I didn't understood the purpose of `start` in this statement.
> > In the event there is no parent node and if it returns an empty `Iterator` calling `.head` will throw error on it.
> 
> @pandurangpatil Won't this lead of `nullPointerException` if there is no parent ?

That's what the null check is being done in the original code.

khemrajrathore: @hiteshbedre Have discussed this with @pandurangpatil and have confirmed the query needs to be 
`node.in(EdgeTypes.AST).collectAll[SqlTableNode].headOption` please make appropriate change and confirm.

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
hiteshbedre: - `TypeDecl` object is required when we are processing any ORM framework.
- `AstNode` object is required when we are processing plain SQL queries(SQLQueryParser).
-  Apart from `buildAndAddSqlQueryNodes` method every other method has complete different implementation.
-  I think we should inherit `BaseORMParser` only in cases of ORM framework processing.
- We should also think of creating parent class for processing plain SQL queries which can be inherited by other languages too.

In Commit: 2d7cb9d5f36f912f5b0a2a71d5b9ea9e233cb263 I have removed inheritance of BaseORMParser from `SQLQueryParser`.

src/main/scala/ai/privado/languageEngine/go/passes/orm/BaseORMParser.scala
hiteshbedre: addressed here: eae285f7239d5e7bd9a6a5a07abf055fd8e443b8

src/main/scala/ai/privado/utility/SQLParser.scala
hiteshbedre: you are correct missed adding `-1`
addressed here: feed56c0872054407657c8cb8bef4a2daf9d98cd

khemrajrathore (approved):
LGTM

PR #822, opened 2023-11-10
Kotlin : Remove Download dependency

khemrajrathore:
We are continuously observing the download dependency part for Kotlin fails leading to scan failure, also for getting types we have the TypeRecovery in place


pandurangpatil (approved):
looks good to me

PR #824, opened 2023-11-13
Policy domains filter cover the TPs APIs

dbMundada:


src/main/scala/ai/privado/policyEngine/PolicyExecutor.scala
```diff
@@ -222,7 +222,20 @@ class PolicyExecutor(
...
+        // Sinks.ThirdParties.API.mediaconvert.awsRegion.amazonaws.com
+        // Sinks.ThirdParties.API.axios.com
+        // ThirdParties.SDK.Sendgrid
+        // ----------------Not covered:
+        // Sinks.API.InternalAPI
+        // Sinks.ThirdParties.API
+        if (sinkId.contains(f"${Constants.thirdPartiesAPIRuleId}.")) {
+          val parts = sinkId.split("\\.")
+          val domain = parts.slice(parts.length - 2, parts.length).mkString(".")
+          sinkFilters.domains.toSet.intersect(Set(domain)).nonEmpty
```
khemrajrathore, line 235: @dbMundada This feels little hacky approach to me, Can we not take `regex` as input for domain, and match it here?

Also if you still want to move ahead with above proposed solution, make this a function, add exception handling and write some unit test around this

pandurangpatil: @dbMundada 
I second the inputs given by the @khemrajrathore to take it as a `regex` input through config folder rules. Also, add respective unit tests for the same.

pandurangpatil (changes requested):

src/main/scala/ai/privado/policyEngine/PolicyExecutor.scala
dbMundada: sure, thanks for input. regex will cover one use case which was partially missing. Updating pr...

khemrajrathore (dismissed):
LGTM, check are still red, once Green can be merged

PR #827, opened 2023-11-14
fix - performance issue due to callIn

khemrajrathore:
PR contains fix for performance issues faced in Kotlin Identifier tagger.

Reason was `callIn` returns duplicate nodes, so need to use `dedup` on top on `callIn` (The root cause can be incorrect AST structure for Kotlin frontend)

On top of it, have cached some queries in Identifier tagger, which helps in optimising the queries further

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/source/IdentifierTagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/source/IdentifierTagger.scala
@@ -26,8 +26,8 @@ package ai.privado.languageEngine.java.tagger.source
 import ai.privado.cache.{RuleCache, TaggerCache}
 import ai.privado.model.{CatLevelOne, Constants, InternalTag, RuleInfo}
 import ai.privado.utility.Utilities.*
-import io.shiftleft.codepropertygraph.generated.nodes.TypeDecl
-import io.shiftleft.codepropertygraph.generated.Cpg
+import io.shiftleft.codepropertygraph.generated.nodes.{Identifier, MethodParameterIn, TypeDecl}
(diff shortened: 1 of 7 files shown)


src/main/scala/ai/privado/languageEngine/java/tagger/source/IdentifierTagger.scala
```diff
@@ -90,59 +97,55 @@ class IdentifierTagger(cpg: Cpg, ruleCache: RuleCache, taggerCache: TaggerCache)
...
-                storeForTag(builder, impactedObject, ruleCache)(Constants.catLevelOne, CatLevelOne.DERIVED_SOURCES.name)
-              }
+        val typeDeclMember = typeDeclValEntry._2
+        val typeDeclVal    = typeDeclValEntry._1.fullName
+        // updating cache
+        taggerCache.addItemToTypeDeclMemberCache(typeDeclVal, ruleInfo.id, typeDeclMember)
+        val typeDeclMemberName = typeDeclMember.name
+        // Have started tagging Parameters as well, as in collection points sometimes there is no referencing Identifier present for a local
+
+        (cachedIdentifier.where(_.typeFullName(typeDeclVal)).l ::: cachedParameter.where(_.typeFullName(typeDeclVal)).l)
```
dbMundada, line 113: Do we need these changes to be done for other languages too? Javascript, Ruby, etc?

khemrajrathore: Not necessarily, only if we need these optimization there. 

We can reflect this to other frontends when we take up the refactoring part to make things common as much as possible, otherwise this is always going to be repeated effort

dbMundada (approved):

PR #834, opened 2023-11-16
add - android permission source detection

khemrajrathore:
Will be adding subsequent test case tomorrow, till then you can perform one round of review


tuxology (approved):
LGTM 🌈

tuxology:
A unit test is failing. Also, maybe add some more unit tests for this later

PR #836, opened 2023-11-17
Support for parsing GO mongo ORM parsing

pandurangpatil:
1. Refactored the ORMParser to use a common query structure to identify the `struct` Types being used with ORM library.
2. Also refactored the unit tests to use common base layer more effectively. 
3. Added support for GO Mongo ORM parsing along with respective unit test for the same.


hiteshbedre (approved):
LGTM

PR #838, opened 2023-11-19
Android layout XML collection tagging

tuxology:
- Goes through the `AndroidLayoutXml` nodes we have made
- Finds the ones that matches source rules and then tags them as collection points


pandurangpatil (approved):
looks good to me

dbMundada (approved):

PR #840, opened 2023-11-21
Android permission tagging for java and JS

tuxology:
This just enables permission tagger for Java and JS. Tested on a few OSS repos to see is tagging is working as intended.


pandurangpatil (approved):
looks good to me

PR #842, opened 2023-11-22
[gosrc2cpg] - Additional collection ORM unit test

ankit-privado:
Added unit test for `Echo`, `Chi`, `Fiber`, `Buffalo` framework


src/test/scala/ai/privado/languageEngine/go/tagger/collection/CollectionTaggerTest.scala
```diff
@@ -91,3 +91,198 @@ class CollectionRestTaggerTest extends CollectionTaggerTestBase {
...
+  }
+
+  "CollectionTagger" should {
+    "Test correct collection tagging" in {
+      cpg.method("allArticles").tag.nameExact("COLLECTION_METHOD_ENDPOINT").head.value shouldBe "\"/articles\""
+    }
+  }
+}
+
+class CollectionTaggerGoBuffaloFrameworkTest extends CollectionTaggerTestBase {
```
pandurangpatil, line 235: @ankit-privado Why we have created a class for each unit test?

ankit-privado: @pandurangpatil For every ORM framework, I have created its own test class.

pandurangpatil: I think we can merge them into one class. But looks like the `private def getContent()` and `goFileContentMap` variables need to be defined separately. 

For now I am approving it. We have to work on a proper solution and merge all of these unit tests inside one class.

pandurangpatil (approved):
looks good to me.

PR #844, opened 2023-11-23
Use android collection rules for tagging

tuxology:
With this approach, we won't use source rules and directly use special collection rules for android forms. This improves tagging coverage since source rules were missing multiple androidXmlTagging node names which are usually very consistent (`emailEditText`, `email_edit_text` etc). This also allows us to add new rules fast and get them added as collections

This has a dependency on android collection rules PR: https://github.com/Privado-Inc/privado/pull/356


pandurangpatil (approved):
Looks good

PR #854, opened 2023-11-30
fix - aws codebuild timeout issue

khemrajrathore:
To fix a test case `fork := true` was added in build.sbt, which is causing failure in aws codebuild

This PR reverts that


pandurangpatil (approved):
looks good

PR #862, opened 2023-12-06
fix android collections source

tuxology:
* fixes android collections so we can view in UI
 * `sourceId` is needed to be added as "tags" in android rules
 * this is needed to link collections to sources
 * also group and dedup collections


khemrajrathore:
Merge this after https://github.com/Privado-Inc/privado/pull/367 gets merged

PR #869, opened 2023-12-07
Few logger fixes

pandurangpatil:
1. Added line number and package of the class where error occured in the log message.
2. Fixed the slf4J error when we run it from IntelliJ IDEA.


khemrajrathore (approved):
LGTM

PR #870, opened 2023-12-11
Fixes android collection FPs

tuxology:
* also adds linked XML filename in code excerpt
 * filter for FP looks for either `R.id` or `binding` in sibling. This might avoid additional `findViewById` filter


khemrajrathore (approved):
LGTM

khemrajrathore:
@tuxology There seems some conflict with base branch

PR #878, opened 2023-12-12
the change is expected with joern changes

pandurangpatil:
the change is expected with joern changes, hence reverting the change.


khemrajrathore (approved):
LGTM

PR #884, opened 2023-12-15
Log4j: Changed console logs

hiteshbedre:
**Context**:
Currently we are populating print statement and error log on console.

**Requirements**:
Logs getting printed on console should work as is and additional to that we need WARN logs from privado-core plus additional few libraries collected in a single file for debugging purpose.

**Changes**:
Console Logs: Addition to print statement we have added ERROR statement will be printed over console: https://github.com/Privado-Inc/privado-core/pull/884/files#diff-618cc4f8ed2a13c1d6cc9405f35994ddf20869ca42eec1438626d5ad9baf06d0R14

Logs collected in file: privado-core's DEBUG level and some other library's WARN level logs are collected: https://github.com/Privado-Inc/privado-core/pull/884/files#diff-618cc4f8ed2a13c1d6cc9405f35994ddf20869ca42eec1438626d5ad9baf06d0R20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,7 +175,7 @@ fabric.properties
 *.class
 
 # Log file
-*.log
+*.logs
 
 # BlueJ files
 *.ctxt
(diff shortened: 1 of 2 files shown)


khemrajrathore:
@hiteshbedre Let's add the `log` folder to gitIgnore

pandurangpatil (approved):
looks good to me.

PR #886, opened 2023-12-19
fix - cookie name for block node

khemrajrathore:


pandurangpatil (changes requested):
Changes looks good. However unit tests are missing

khemrajrathore:
> Changes looks good. However unit tests are missing

@pandurangpatil Have added test case

pandurangpatil (approved):

PR #900, opened 2024-01-02
Link fieldAccess calls of annotated methods to original properties via member

karan-batavia:
This PR aims to link fieldAccess nodes found in a Spring annotated method, to it's original propertyNode through a member node. Also, the respective language engine version required for this change has been updated.


src/test/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePassTest.scala
```diff
@@ -51,41 +52,49 @@ class AnnotationTests extends PropertiesFilePassTestBase(".properties") {
...
+      cpg.property.usedAt.originalProperty.l.length shouldBe 3
+      cpg.property.usedAt.originalProperty.name.toSet.l shouldBe List("internal.logger.api.base", "slack.base.url")
+      cpg.property.usedAt.originalProperty.value.toSet.l shouldBe List(
         "https://logger.privado.ai/",
         "https://hooks.slack.com/services/some/leaking/url"
       )
     }
+
+    "connect the referenced member to the original property denoted by the annotated method" in {
+      cpg.member("loggerUrl").originalProperty.l.name.head shouldBe "internal.logger.api.base"
```
khemrajrathore, line 95: Instead of checking on head, let's check on the entire list and also verify that there is only 1 original property for this member, similarly for value

src/test/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePassTest.scala
```diff
@@ -51,41 +52,49 @@ class AnnotationTests extends PropertiesFilePassTestBase(".properties") {
...
+        "java.lang.String slackWebHookURL"
+      )
     }
 
     "connect property to annotated parameter" in {
-      cpg.property.usedAt.originalProperty.l.length shouldBe 2
-      cpg.property.usedAt.originalProperty.name.l shouldBe List("internal.logger.api.base", "slack.base.url")
-      cpg.property.usedAt.originalProperty.value.l shouldBe List(
+      cpg.property.usedAt.originalProperty.l.length shouldBe 3
+      cpg.property.usedAt.originalProperty.name.toSet.l shouldBe List("internal.logger.api.base", "slack.base.url")
```
khemrajrathore, line 87: Instead of Set use List to compare the names and values

src/main/scala/ai/privado/languageEngine/java/passes/config/JavaPropertyLinkerPass.scala
```diff
@@ -70,17 +72,20 @@ class JavaPropertyLinkerPass(cpg: Cpg) extends PrivadoParallelCpgPass[JavaProper
...
-      .filter { case (key, _) =>
-        propertyNode.name == Option(key.code.slice(3, key.code.length - 2)).getOrElse("")
-      }
-      .foreach { case (_, value) =>
-        builder.addEdge(propertyNode, value, EdgeTypes.IS_USED_AT)
-        builder.addEdge(value, propertyNode, EdgeTypes.ORIGINAL_PROPERTY)
+    annotatedMethods()
+      .filter { case (key, _) => propertyNode.name == Option(key.code.slice(3, key.code.length - 2)).getOrElse("") }
+      .foreach { case (_, method) =>
+        val referencedMember = method.ast.fieldAccess.referencedMember.l.headOption.orNull
```
khemrajrathore, line 78: There can be multiple field access which may exist in a method, how can we guarantee the head is the one this property was used at?

karan-batavia: These methods are most commonly setters used to inject property values, and set them to a member. Almost all of them are one liners.

khemrajrathore: Agree, but still I see some scope. Can you add a todo stating where this may not work, for future reference?

pandurangpatil: @karan-batavia Ideally you should taint the method parameter. Which in turn will taint the right field access.

pandurangpatil: @karan-batavia sorry for the confusion. What you are doing should work.

karan-batavia:
@khemrajrathore , this is tested and working. Can we merge so this gets released?

khemrajrathore (approved):
LGTM

PR #914, opened 2024-01-05
Get S3 bucket names from code

tuxology:
We try to get S3 bucket names from code and populate them in `Amazon S3` sinks. This PR supports following:

✅  For Python, if you use `boto3` client and use a variable as argument, we will try to see if a bucket literal was directly assigned to the argument somewhere in code
✅ For Python, we also support getting bucket names if `s3a://` urls are used (sometimes used with spark)
✅ For Java, we support official AWS Sdk and see if a bucket name as literal is directly used by client
❌ If a bucket name is within complex structures defined in code (eg. a list of tuples/objects), we don't support it. Only direct assignment to bucket name literal is checked
❌ We don't support bucket names defined externally in configurations/yaml/property files as of now
❌ We map to existing DatabaseDetails structure with some hacks. This will eventually be replaced for ease of use


src/main/scala/ai/privado/tagger/S3Tagger.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.slf4j.LoggerFactory
+import io.shiftleft.codepropertygraph.generated.nodes.{Call, Method}
+import io.shiftleft.semanticcpg.language.*
+import io.shiftleft.codepropertygraph.generated.nodes.JavaProperty
+import org.slf4j.LoggerFactory
+import overflowdb.traversal.*
+
+import scala.collection.mutable.ListBuffer
+
+class S3Tagger(cpg: Cpg) extends PrivadoSimpleCpgPass(cpg) {
```
khemrajrathore, line 16: Any reason to use PrivadoSimpleCpgPass instead of Concurrent pass?

src/main/scala/ai/privado/tagger/S3Tagger.scala
```diff
@@ -0,0 +1,130 @@
...
+  private val readRuleId      = "Storages.AmazonS3.Read"
+  private val writeRuleId     = "Storages.AmazonS3.Write"
+  private val readWriteRuleId = "Storages.AmazonS3.ReadAndWrite"
+
+  override def run(builder: DiffGraphBuilder): Unit = {
+    val readBucketNames      = ListBuffer[String]()
+    val writeBucketNames     = ListBuffer[String]()
+    val readWriteBucketNames = ListBuffer[String]()
+
+    if (cpg.metaData.language.head == Languages.JAVASRC) {
```
khemrajrathore, line 29: I see we are using conditions, can we not have an abstract S3BaseTagger, whose `generateParts` functions can be defined at the language level.

Effectively having `JavaS3Tagger` and `PythonS3Tagger` extending `S3BaseTagger`

src/main/scala/ai/privado/tagger/S3Tagger.scala
```diff
@@ -0,0 +1,130 @@
...
+            bucket.stripPrefix("\"").stripSuffix("\"").stripPrefix("\'").stripSuffix("\'").stripPrefix("s3a://")
+          if bucketName.matches(bucketPattern) then bucketName
+          else ""
+        })
+        .filter(_.nonEmpty)
+        .l
+    }
+
+    // remove existing DB details from rule and add READ DB DETAILS to separate S3 DB details
+    DatabaseDetailsCache.removeDatabaseDetails(readRuleId)
```
khemrajrathore, line 107: Any reason to remove this, do we suspect this getting added in DatabaseDetailsCache?

src/main/scala/ai/privado/cache/S3DatabaseDetailsCache.scala
```diff
@@ -0,0 +1,39 @@
...
+
+package ai.privado.cache
+
+import ai.privado.model.DatabaseDetails
+
+import scala.collection.mutable
+
+/** Cache to store Rules specific things
+  */
+object S3DatabaseDetailsCache {
```
khemrajrathore, line 31: Instead of using `object` make this an `class` so that we can use this as an instance, this help in solving overlapping test case issue

khemrajrathore:
- [ ] Please Add Test cases for covered case, and if possible ignored test case remaining case

src/main/scala/ai/privado/tagger/S3Tagger.scala
tuxology: There is no clear parallelization we can do there I think 🤔 There are only three types of rules with different logic that runs linearly for them. We can think about optimizing this later IMO

src/main/scala/ai/privado/tagger/S3Tagger.scala
tuxology: Ok, yes, this is a way. It will also avoid needing to run the tagger for each language as I am doing now

tuxology: Since this will eventually be overhauled and added to a totally new system, can we let this in? We see S3 tagging used in its current state for half a month more before we restructure this and move this in a more comprehensive asset tagging

src/main/scala/ai/privado/tagger/S3Tagger.scala
tuxology: Yes, since tagging has already run before, the DatabaseDetails already gets populated in the cache. We need to remove it once and add new info. This pass is run after source tagging is already done.

src/main/scala/ai/privado/cache/S3DatabaseDetailsCache.scala
tuxology: This can be done, yes 👍

src/main/scala/ai/privado/tagger/S3Tagger.scala
khemrajrathore: Better to do it now, as things move slowly once in backlog

pandurangpatil (approved):
looks good to me

PR #912, opened 2024-01-04
Add support for PHP

pawan-privado:
This PR checks in the processor for `PHP` as part of support for `PHP` language.

PR #915, opened 2024-01-07
Check in taggers for PHP

pawan-privado:
This PR checks in the taggers required for supporting PHP. In addition to that, based on my discussions with Khemraj, I've moved the `IdentifierTagger` from `ruby` package to the common package for its reuse.

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/php/tagger/PrivadoTagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/php/tagger/PrivadoTagger.scala
@@ -25,10 +25,9 @@ package ai.privado.languageEngine.php.tagger
 
 import ai.privado.cache.{DataFlowCache, RuleCache, TaggerCache}
 import ai.privado.entrypoint.PrivadoInput
-import ai.privado.languageEngine.java.tagger.source.IdentifierTagger
 import ai.privado.tagger.PrivadoBaseTagger
 import ai.privado.tagger.sink.RegularSinkTagger
-import ai.privado.tagger.source.LiteralTagger
+import ai.privado.tagger.source.{IdentifierTagger, LiteralTagger}
 import io.shiftleft.codepropertygraph.generated.Cpg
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/privado/languageEngine/php/tagger/PrivadoTagger.scala
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * For more information, contact support@privado.ai
+ *
+ */
+
+package ai.privado.languageEngine.php.tagger
+
+import ai.privado.cache.{DataFlowCache, RuleCache, TaggerCache}
+import ai.privado.entrypoint.PrivadoInput
+import ai.privado.languageEngine.java.tagger.source.IdentifierTagger
```
khemrajrathore, line 28: I see Java's Identifier tagger being used here

src/main/scala/ai/privado/tagger/source/IdentifierTagger.scala
```diff
@@ -21,7 +21,7 @@
  *
  */
 
-package ai.privado.languageEngine.ruby.tagger.source
+package ai.privado.tagger.source
```
khemrajrathore, line 24: Did we only replaced the package name, ideally we should also do a refactor to place this in the appropriate package as well

pawan-privado: Yes, I've placed it in the appropriate package and GH is indicating that the file has been moved to a new location. Or am I mistaken here?

src/main/scala/ai/privado/languageEngine/php/tagger/PrivadoTagger.scala
pawan-privado: Fixed. My bad!

pawan-privado:
Moved changes to #912 as requested.

PR #926, opened 2024-01-23
ingress: added ingress and egress

hiteshbedre:
Requirements:
1. Collect all ingress and egress urls.
2. Add those entries in privado.json
3. Add flag for above, default it would be off.

### DIFF
--- a/src/main/scala/ai/privado/exporter/PropertyExporter.scala
+++ b/src/main/scala/ai/privado/exporter/PropertyExporter.scala
@@ -39,25 +39,25 @@ class PropertyExporter(cpg: Cpg, ruleCache: RuleCache) {
   private val SPRING_ANNOTATION_ID = "Collections.Annotation.Spring"
   private val REGEX_SLASH          = ".*/.*"
 
-  def getPropertyUrls = {
-    var propertyUrls = List[String]()
+  def getEgressUrls = {
+    var egressUrls = List[String]()
     val apiRules = ruleCache.getAllRuleInfo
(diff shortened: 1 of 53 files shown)


src/main/scala/ai/privado/exporter/PropertyExporter.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    egressUrls.concat(addUrlFromFeignClient())
+    egressUrls.dedup.l
+  }
+
+  private def addUrlFromFeignClient(): List[String] = {
+
+    var egressUrls = List[String]()
+    try {
+      val filesHavingFeignClient = cpg.annotation.name(FEIGN_CLIENT).file.name.l
```
hiteshbedre, line 62: For annotation used in `interface` we are able to get its `typeDecl` node thats why I have created a `filesHavingFeignClient` cache to filter out non required annotation.

Query I used to came to conclusion:
cpg.annotation.name("<some_annotation_name>").typeDecl.size => 0

Note: When a annotation is defined over a method under a class then we are able to trace its `typeDecl` node, but in case of `interface` its not.

pandurangpatil (changes requested):
Add unit test ingress

hiteshbedre:
> Add unit test ingress

Added here: https://github.com/Privado-Inc/privado-core/pull/926/commits/7b658e265371c7b249e114a324e59500f3ed9825

pandurangpatil (approved):

pandurangpatil (approved):
Looks good

PR #936, opened 2024-01-30
changed log handling and exit handling

pandurangpatil:
1. Made changes in few log statments to log the complete stack trace.
2. Made changes to convert debug to trace logs for rule valiation.
3. Added exit(1) in main error handling situation, to indicate outer process that the scan failed. Made respective change in matric collection method calls as `finally` block will not be invoked after `exit(1)` call.


ojaswa1942 (approved):
LGTM!

PR #938, opened 2024-02-01
add asset discovery flag

khemrajrathore:
- add new flag `asset-discovery`
- skip reading `.privado` folder when running Property Pass


hiteshbedre (approved):
LGTM

PR #944, opened 2024-02-05
added pattern for egress

hiteshbedre:
Pattern for Egress:
1. Look into property file.
2. Add values starts with slash or value contain more than two slash.

Property pass for JS repos


pandurangpatil (changes requested):
@hiteshbedre missing unit test

pandurangpatil (approved):
Looks good

PR #947, opened 2024-02-06
add - support for reading creating property nodes from .js, .py file

khemrajrathore:
As properties can be defined in .js, .py file, this PR focusses on creating nodes for them


pandurangpatil (approved):
Looks good

PR #952, opened 2024-02-08
Add/asset discovery on generated json

khemrajrathore:


src/main/scala/ai/privado/exporter/ExporterUtility.scala
```diff
@@ -405,17 +405,25 @@ object ExporterUtility {
...
+              "(?i).*(connection|host|database|location|uri|mongo|sql|postgres|s3|oracle|redis|bucket|dynamo|maria|db(_)?name).*"
+            )
+          ),
+          _.filter(_._2.matches("(?i).*(mongo|sql|postgres|aws|oracle|redis|dynamo|maria).*"))
         )
         .filterNot(_._2.matches(".*[.](png|jpg|jpeg|jar|zip|xml|json|yml)$"))
         .filterNot(_._2.matches("^(true|false)$"))
         .l
 
+      println("Printing probable assets")
```
hiteshbedre, line 419: is this required to be printed over console?

khemrajrathore: Yes, for debugging. This is only be printed in case we pass flag, which we will be doing only for POC purpose

hiteshbedre (approved):
LGTM

PR #956, opened 2024-02-08
Changed JS API tagger

pandurangpatil:
Changed JS API tagger to run the queries on `cpg.code` instead of methodFullName

PR #957, opened 2024-02-08
Changed JS API tagger (#956)

pandurangpatil:
* Changed JS API tagger

Changed JS API tagger to run the queries on `cpg.code` instead of methodFullName

* including both method full name and code

* optimised the OR query


khemrajrathore (approved):
LGTM

PR #960, opened 2024-02-14
Add/asset discovery on generated json (#952)

khemrajrathore:
* generate property nodes from generated jsons

* scan property from .privado/config folder

* fix - json parser bug


hiteshbedre (approved):
LGTM

PR #964, opened 2024-02-16
Express cp

khemrajrathore:


src/main/scala/ai/privado/languageEngine/javascript/tagger/collection/CollectionTagger.scala
```diff
@@ -136,6 +143,45 @@ class CollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelCp
...
+
+     */
+
+    callMap.foreach(callEntry => {
+
+      methodUrlMap.get(callEntry._1) match
+        case Some(endPoint) =>
+          val params = endPoint
+            .stripPrefix("\"")
+            .stripPrefix("'")
```
HiteshMah-Jan, line 165: could be double quotes or backticks as well. we can just strip any non alpha character here ?

khemrajrathore: I guess, if its a backticks that is already handled when adding to `methodUrlMap`

But anyways let me replace nonAlpha characters

PR #966, opened 2024-02-19
improve regex

khemrajrathore:


src/main/scala/ai/privado/exporter/ExporterUtility.scala
```diff
@@ -404,16 +404,18 @@ object ExporterUtility {
...
             _._1.matches(
-              "(?i).*(connection|host|database|location|uri|mongo|sql|postgres|s3|oracle|redis|bucket|dynamo|maria|db(_)?name).*"
+              "(?i).*(connection|host|database|location|uri$|mongo|sql|postgres|s3|oracle|redis|bucket|dynamo|maria|db(_)?name).*"
             )
           ),
           _.filter(_._2.matches("(?i).*(mongo|sql|postgres|aws|oracle|redis|dynamo|maria).*"))
         )
         .filterNot(_._2.matches(".*[.](png|jpg|jpeg|jar|zip|xml|json|yml)$"))
         .filterNot(_._2.matches("^(true|false)$"))
+        .filterNot(_._2.matches("regex[(]"))
```
hiteshbedre, line 418: Can you share a example where you want to ignore values having `regex` as word in it.

khemrajrathore: Good catch @hiteshbedre, it should be `require` instead of regex

PR #976, opened 2024-02-23
fix - lombok folder deletion

khemrajrathore:
PR fixes the following
- The check to delete `delombok` folder once scan is completed was commented out earlier while testing, which was released. Have uncommented that now, this fixes the deletion of delombok code once scan is completed succesfully
- Earlier we were creating a `delombok` directory and then invoking the delombok generator which was generating 2 delombok folder. Have removed the logic to create a `delombok` directory ourself. This fixes the creation of 2 delombok folder instead of the expected 1 folder


HiteshMah-Jan (approved):
Looks good

PR #979, opened 2024-02-28
Product Level Datamap: Application base path

hiteshbedre:
Requirement:
1. Gather application base path if exist.
2. Base Path: Root resource of microservice.

Changes:
1. Renamed `EgressExporter` to `HttpConnectionMetadataExporter`
2. Fulfilled requirements mentioned above.
3. Updated test case.


khemrajrathore (approved):
LGTM

khemrajrathore:
@hiteshbedre Just to highlight, when we pass the `asset_discovery` flag all the config files are read irrespective of language which is consumed by `privado-core` and corresponding `propertyNodes` are created. Maybe in future if we plan to get the base path for non java language that can be used
cc- @nikvin15

PR #983, opened 2024-02-29
PHP Integration

tuxology:
* integrate PHP Frontend from upstream Joern into Privado
* vendor PHP Parser from upstream fork into Privado and use it 
* Use 4GB heap size for the tests to run in CI.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,5 @@
 import sbt.Credentials
+import better.files.File
 
 name                     := "privado-core"
 ThisBuild / organization := "ai.privado"
@@ -149,6 +150,8 @@ goAstGenDlTask := {
   distDir.listFiles().foreach(_.setExecutable(true, false))
 }
 
+Compile / compile := (Compile / compile).dependsOn(goAstGenDlTask, phpParseDlTask).value
+
 lazy val goAstGenSetAllPlatforms = taskKey[Unit](s"Set ALL_PLATFORMS")
 goAstGenSetAllPlatforms := { System.setProperty("ALL_PLATFORMS", "TRUE") }
 
@@ -206,12 +209,40 @@ stage := Def
 
 Compile / compile := ((Compile / compile) dependsOn dotnetAstGenDlTask).value
 
-// Also remove astgen binaries with clean, e.g., to allow for updating them.
+// Download php-parser: start
+// This is based on how php2cpg vendors the php-parser in Joern
+val phpParserVersion       = "4.15.7"
+val upstreamParserBinName  = "php-parser.phar"
+val versionedParserBinName = s"php-parser-$phpParserVersion.phar"
+val phpParserDlUrl =
+  s"https://github.com/joernio/PHP-Parser/releases/download/v$phpParserVersion/$upstreamParserBinName"
+
+lazy val phpParseDlTask = taskKey[Unit]("Download php-parser binaries")
+phpParseDlTask := {
+  val phpBinDir = baseDirectory.value / "bin" / "php-parser"
+  phpBinDir.mkdirs()
+
+  val downloadedFile = SimpleCache.downloadMaybe(phpParserDlUrl)
+  IO.copyFile(downloadedFile, phpBinDir / upstreamParserBinName)
+
+  File((phpBinDir / "php-parser.php").getPath)
+    .createFileIfNotExists()
+    .overwrite(s"<?php\nrequire('$versionedParserBinName');?>")
+
+  val distDir = (Universal / stagingDirectory).value / "bin" / "php-parser"
+  distDir.mkdirs()
+  IO.copyDirectory(phpBinDir, distDir)
+}
+// Download php-parser: end
+
+// Also remove astgen and php-parser binaries with clean, e.g., to allow for updating them.
 // Sadly, we can't define the bin/ folders globally,
 // as .value can only be used within a task or setting macro
 cleanFiles ++= Seq(
   baseDirectory.value / "bin" / "astgen",
-  (Universal / stagingDirectory).value / "bin" / "astgen"
+  (Universal / stagingDirectory).value / "bin" / "astgen",
+  baseDirectory.value / "bin" / "php-parser",
+  (Universal / stagingDirectory).value / "bin" / "php-parser"
 ) ++ astGenBinaryNames.map(fileName => SimpleCache.encodeFile(s"$astGenDlUrl$fileName"))
 Compile / doc / sources                := Seq.empty
 Compile / packageDoc / publishArtifact := false
--- a/build.sbt
+++ b/build.sbt
@@ -150,8 +150,6 @@ goAstGenDlTask := {
   distDir.listFiles().foreach(_.setExecutable(true, false))
 }
 
-Compile / compile := (Compile / compile).dependsOn(goAstGenDlTask, phpParseDlTask).value
-
 lazy val goAstGenSetAllPlatforms = taskKey[Unit](s"Set ALL_PLATFORMS")
 goAstGenSetAllPlatforms := { System.setProperty("ALL_PLATFORMS", "TRUE") }
 
@@ -216,6 +214,8 @@ val upstreamParserBinName  = "php-parser.phar"
 val versionedParserBinName = s"php-parser-$phpParserVersion.phar"
 val phpParserDlUrl =
   s"https://github.com/joernio/PHP-Parser/releases/download/v$phpParserVersion/$upstreamParserBinName"
+
+Compile / compile := ((Compile / compile) dependsOn phpParseDlTask).value
 
 lazy val phpParseDlTask = taskKey[Unit]("Download php-parser binaries")
 phpParseDlTask := {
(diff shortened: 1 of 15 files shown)


khemrajrathore (commented):
Overall changes look good to me, If we can also have some test cases around PHP parsing and tagging that would be great

karan-batavia:
Can we please have some tests on the basic tagging process? Let's ensure the integration stays solid from the start itself.

pawan-privado:
> Can we please have some tests on the basic tagging process? Let's ensure the integration stays solid from the start itself. @pawan-privado

Sure!

karan-batavia:
Also, When are we planning to release these changes? Let's make sure we don't conflict with the C# release on Monday, 
@pawan-privado @tuxology

build.sbt
```diff
@@ -147,7 +148,7 @@ goAstGenDlTask := {
   distDir.listFiles().foreach(_.setExecutable(true, false))
 }
 
-Compile / compile := ((Compile / compile) dependsOn goAstGenDlTask).value
+Compile / compile := (Compile / compile).dependsOn(goAstGenDlTask, phpParseDlTask).value
```
karan-batavia, line 151: This doesn't work on Linux, can you please test once and confirm? @pawan-privado

pawan-privado: I don't have access to a Linux machine ATM but it works fine on macOS. I've DM'd you something on Slack, btw, regarding this.

pawan-privado:
@tuxology Test failed because PHP parser could not be executed.
```
Error:  Invalid path for PhpParserBin: /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/io/joern/php2cpg/bin/php-parser/php-parser.php
```

The path doesn't look right to me. Is this the intended path for vendoring the parser?

build.sbt
tuxology: This looked fine to me - I tested on Linux, but did a cleaner version of this just in case

tuxology:
> @tuxology Test failed because PHP parser could not be executed.
> 
> ```
> Error:  Invalid path for PhpParserBin: /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/io/joern/php2cpg/bin/php-parser/php-parser.php
> ```
> 
> The path doesn't look right to me. Is this the intended path for vendoring the parser?

@pawan-privado It looks like this is only happening in Github - the path for `php-parser` is possibly null and hence the default path is being picked (which is not vendored path). This was working before C# push (https://github.com/Privado-Inc/privado-core/pull/983/commits/a883357d7bdf73735da07bec602850f1454446ae) so if you can investigate and fix, it would help 👍

src/main/scala/ai/privado/languageEngine/php/processor/PhpProcessor.scala
```diff
@@ -0,0 +1,107 @@
...
+        new File(dir.substring("file:".length, indexOfLib)).toString
+      } else {
+        val indexOfTarget = dir.lastIndexOf("target")
+        if (indexOfTarget != -1) {
+          new File(dir.substring("file:".length, indexOfTarget)).toString
+        } else {
+          "."
+        }
+      }
+      Paths.get(fixedDir, "/bin/php-parser/php-parser.php").toAbsolutePath.toString
```
pawan-privado, line 87: 1. This path looks suspiciously wrong. I have a `bin` folder in my parser which has a single file called `php-parse`.
2. You need to run `composer install` prior to using the parser.

pawan-privado:
> > @tuxology Test failed because PHP parser could not be executed.
> > ```
> > Error:  Invalid path for PhpParserBin: /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/io/joern/php2cpg/bin/php-parser/php-parser.php
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The path doesn't look right to me. Is this the intended path for vendoring the parser?
> 
> @pawan-privado It looks like this is only happening in Github - the path for `php-parser` is possibly null and hence the default path is being picked (which is not vendored path). This was working before C# push ([a883357](https://github.com/Privado-Inc/privado-core/commit/a883357d7bdf73735da07bec602850f1454446ae)) so if you can investigate and fix, it would help 👍

The reason why we did not notice this happen in our dev machines is because we had the required `env` variable set and the path was being picked up from the variable.

build.sbt
pawan-privado: > This doesn't work on Linux, can you please test once and confirm? @pawan-privado

Haven't noticed any abnormalities at my end. Can you attach the error log if it still continues to happen? AFAIK, there's no platform specific stage in our build process so it shouldn't fail on Linux systems.

src/test/scala/ai/privado/languageEngine/php/tagger/source/FieldIdentifierTaggingTests.scala
```diff
@@ -0,0 +1,52 @@
...
+          |
+          |  function add_phone($ph) {
+          |    //
+          |  }
+          |}
+          |?>
+          |""".stripMargin)
+
+      val List(firstName) = cpg.member.l
+      firstName.tag.nameExact(Constants.catLevelOne).value.l shouldBe List(CatLevelOne.SOURCES.name)
```
karan-batavia, line 49: Are we testing field identifiers here, or members? I don't see a field access node being accessed.

karan-batavia: You can refer C# Field Identifier Tests

pawan-privado: Fixed.

src/test/scala/ai/privado/languageEngine/php/tagger/source/IdentifierTaggingTest.scala
```diff
@@ -0,0 +1,62 @@
...
+        |  function get_email() {
+        |    return $emailId;
+        |  }
+        |}
+        |?>
+        |
+        |""".stripMargin)
+
+    "tag member in a structure" in {
+      cpg.member("firstName").tag.nameExact(Constants.id).value.l shouldBe List(
```
karan-batavia, line 53: Let's test for identifiers as well here.

src/test/scala/ai/privado/languageEngine/php/tagger/source/IdentifierTaggingTest.scala
```diff
@@ -0,0 +1,62 @@
...
+        "Data.Sensitive.PersonalIdentification.FirstName"
+      )
+
+      cpg.member("dob").tag.nameExact(Constants.id).value.l shouldBe List(
```
karan-batavia, line 57: Here as well.

kbatavia12 (approved):
LGTM

karan-batavia (dismissed):

karan-batavia (dismissed):

src/test/scala/ai/privado/languageEngine/php/PhpTestBase.scala
```diff
@@ -0,0 +1,181 @@
...
+import org.scalatest.{BeforeAndAfterAll, BeforeAndAfterEach}
+
+import scala.collection.mutable
+
+abstract class PhpTestBase extends AnyWordSpec with Matchers with BeforeAndAfterAll with BeforeAndAfterEach {
+  private val cpgs        = mutable.ArrayBuffer.empty[Cpg]
+  private val outPutFiles = mutable.ArrayBuffer.empty[File]
+  private val inputDirs   = mutable.ArrayBuffer.empty[File]
+
+  private val sourceRules = List(
```
khemrajrathore, line 48: The test source rules are defined at https://github.com/Privado-Inc/privado-core/blob/main/src/test/scala/ai/privado/RuleInfoTestData.scala kindly use that

src/test/scala/ai/privado/languageEngine/php/tagger/source/FieldIdentifierTaggingTests.scala
```diff
@@ -0,0 +1,53 @@
...
+          |
+          |  function add_phone($ph) {
+          |    //
+          |  }
+          |}
+          |?>
+          |""".stripMargin)
+
+      val List(firstNameField) = cpg.fieldAccess.l
+      firstNameField.tag.nameExact(Constants.catLevelOne).value.l shouldBe List(CatLevelOne.SOURCES.name)
```
khemrajrathore, line 50: also add a check on the `code` of `firstNameField` to assert we are getting the expected fieldAccess tagged

src/test/scala/ai/privado/languageEngine/php/tagger/source/LiteralTaggingTests.scala
```diff
@@ -0,0 +1,49 @@
...
+          |  }
+          |
+          |  function add_phone($ph) {
+          |  }
+          |}
+          |?>
+          |""".stripMargin)
+
+      val literals = cpg.literal.l
+      literals.last.tag.nameExact(Constants.catLevelOne).value.l shouldBe List(CatLevelOne.SOURCES.name)
```
khemrajrathore, line 46: Also add a check on the `code` of literal to assert we are tagging the right literal

khemrajrathore:
@pawan-privado Take a merge of dev, they are are good to go

khemrajrathore (approved):
LGTM

PR #993, opened 2024-03-05
Fix stage error for `DotNetAstGen`

karan-batavia:
The production build works fine, `sbt clean stage` fails.


pandurangpatil (approved):
Looks good

PR #994, opened 2024-03-06
Fix Extraction of URL from Request Mapping Annotations

hiteshbedre:
**Description:**
This pull request addresses a critical issue where the actual URL of a collection is not correctly extracted under specific conditions. The bug manifests when the first parameter of annotations such as  @RequestMapping ,  @GetMapping ,  @PostMapping , etc., is something other than the intended URL. Previously, our logic incorrectly assumed that the first parameter always represented the URL. 

**Key Changes:** 
 
1. **Issue Identification:** Identified a breaking case where the extraction logic fails if the first parameter of the mapping annotation does not directly represent the URL. 
 
2. **Existing Logic Revision:** The existing extraction logic was limited to evaluating only the first parameter of the annotation, which led to inaccuracies in URL retrieval. 
 
3. **Focused Extraction:** Revised the logic to prioritize the extraction of the  value  parameter specifically, as it directly corresponds to the intended URL. In cases where the  value  parameter is not explicitly named, the logic defaults to extracting the direct URL string provided. 
 
4. **CPG Query Update:** Modified the code property graph (CPG) query to accurately extract the value associated with the  value  parameter, or to proceed without a parameter when a direct URL string is provided. 
 
5. **Testing and Verification:** Conducted thorough testing to ensure the accuracy of URL extraction across all scenarios. This included verifying all existing test cases related to collection URLs and introducing tests for the identified breaking case to prevent future regressions.

### DIFF
--- a/src/test/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtilityTest.scala
+++ b/src/test/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtilityTest.scala
@@ -58,6 +58,9 @@ class CollectionUtilityTest extends JavaTaggingTestBase {
       |	public Token login(@RequestBody String somestring) {
       | }
       |
+      | @RequestMapping(method = RequestMethod.GET, value = "/products", produces = "application/json")
+      | public List<Product> getProducts() {
+      |    }
       |}
       |
       |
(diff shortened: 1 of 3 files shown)


src/test/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtilityTest.scala
```diff
@@ -58,6 +58,9 @@ class CollectionUtilityTest extends JavaTaggingTestBase {
       |	public Token login(@RequestBody String somestring) {
       | }
       |
+      | @RequestMapping(method = RequestMethod.GET, value = "/products", produces = "application/json")
```
khemrajrathore, line 61: @hiteshbedre Please also add a test case for `@PostMapping("/products")`, I believe above implementation will fail for this case

khemrajrathore: never mind, that is already handled in `none`

PR #998, opened 2024-03-07
LE-1096 | Add method full name collection tagger to Kotlin tagger

sriram-privado:
Add missing method full name tagger to the Kotlin tagger - missed out before, was only added to the Java tagger.


sriram-privado:
Sorry, long day, this was already there in the Kotlin tagger.

PR #999, opened 2024-03-08
added base path for feign client

hiteshbedre:
##### Example:
```
@FeignClient(name = "address-service", url = "http://localhost:8081", path = "/address-service")
public interface AddressClient {

	@GetMapping(value = "/address/{id}")
	public ResponseEntity<AddressResponse> getAddressByEmployeeId(@PathVariable("id") int id);

}
```
Requirement:
* Add base path as part of url.
      - url generating before: **/address/{id}**
      - url generating after change: **address-service/address/{id}**

Updated and verified existing test cases.

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtility.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtility.scala
@@ -160,7 +160,7 @@ object CollectionUtility {
     * @return
     */
   def getUrlFromAnnotation(annotation: Annotation): String = {
-    annotation.parameterAssign.where(_.parameter.code("value")).value.l.headOption match {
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/privado/languageEngine/java/tagger/collection/CollectionUtility.scala
```diff
@@ -160,7 +160,7 @@ object CollectionUtility {
     * @return
     */
   def getUrlFromAnnotation(annotation: Annotation): String = {
-    annotation.parameterAssign.where(_.parameter.code("value")).value.l.headOption match {
+    annotation.parameterAssign.where(_.parameter.code("value|name")).value.l.headOption match {
```
khemrajrathore, line 163: I believe in this case it should be picked up from the property `path` not `name`

khemrajrathore (approved):
LGTM

PR #1000, opened 2024-03-08
Add support to collect egress from code for js, python and java

nikvin15:
### Collect egresses from code for js, python and java
- Cases Handled
    -  String concatenation using addition operator 
    -  String formatter operator
    - Literal without any operation and getting assigned to a variable. 
- Implementation Details
   - Collecting all literals that has slash in it and allowing some specific character to also support string formatter
   - Filtering if only non alphanumeric symbols present
   - remove \\" from all the strings
   - go to call node(_we are not handling nested level concatination/formatting for now_) and collect strings based on the call node and their arguments
   
- The current logic is designed for Java, Python, and JavaScript, as I have tested it for these languages. I have not extended it to other languages yet, as literals can also appear in imports. In such cases, I will need to write additional logic to handle this scenario. This was done for javascript.

### DIFF
--- a/src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
+++ b/src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
@@ -28,7 +28,7 @@ import ai.privado.semantic.Language.*
 import ai.privado.languageEngine.java.language.*
 import io.shiftleft.codepropertygraph.generated.Cpg
 import org.slf4j.LoggerFactory
-import ai.privado.model.{Constants, Language, NodeType}
+import ai.privado.model.{Constants, Language}
 import io.shiftleft.semanticcpg.language.*
 import ai.privado.languageEngine.java.tagger.collection.CollectionUtility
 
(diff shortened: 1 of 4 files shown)


src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
```diff
@@ -41,22 +41,63 @@ class HttpConnectionMetadataExporter(cpg: Cpg, ruleCache: RuleCache) {
...
+  private val IMPORT_REGEX_WITH_SLASHES                            = "(?s)^(?=.*/)(?!.*/$).*"
 
   private val LAMBDA_SERVERLESS_BASE_PATH = "service"
   private val LAMBDA_SERVERLESS_FILE_NAME = ".*serverless.yml"
 
-  def getEgressUrls = {
+  def getLiteralsFromLanguageFiles: List[String] = {
+    val egressLiterals = cpg
+      .literal(URL_PATH_WITH_VARIABLE_SYMBOLS)
+      .filter(node => node.code.replaceAll(NON_ALPHANUMERIC, "") != "")
```
khemrajrathore, line 56: Can we not check here, if the code is `ALPHANUMERIC`, instead of replacing it?

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
```diff
@@ -41,22 +41,63 @@ class HttpConnectionMetadataExporter(cpg: Cpg, ruleCache: RuleCache) {
...
+      .filter(!_.code.matches(STRING_WITH_CONSECUTIVE_DOTS_OR_DOT_SLASH_OR_NEWLINE))
+      .inCall
+      .map(node =>
+        node.argument
+          .map(arg => {
+            if node.name == "require" then ""   // skip import statement
+            else if arg.isLiteral then arg.code // collect literal
+            // const loginPath = "api/v1" + "/login" --- Addition Case(javascript, python, java)
+            // const signupPath = `api/v1/${signup}` --- Format String Case for javascript(similar applicable for python, java)
+            else if node.name.indexOf("addition") > -1 || node.name.indexOf("formatString") > -1 then arg.code
```
khemrajrathore, line 66: I guess here you are checking if the call node is a `addition` or `formatString`, an optimized way would be
`node.name.equals("<operator>.addition")` and similarly for formatString. Because I believe the time complexity of `equals` is better than `indexOf`

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
```diff
@@ -41,22 +41,63 @@ class HttpConnectionMetadataExporter(cpg: Cpg, ruleCache: RuleCache) {
...
     var egressUrls = List[String]()
 
     egressUrls = egressUrls.concat(
       cpg.property.or(_.value(STRING_START_WITH_SLASH), _.value(STRING_CONTAINS_TWO_SLASH)).value.dedup.l
     )
+    /* We have verified literals for these languages, so we need to analyze other languages before broadening the rule.
+       It can happen that literals may come from imports as well, which was the case for JavaScript, and we handled it.
+       Therefore, we might need to do a few things specific to each language. */
+    if (
+      AppCache.repoLanguage.id == Language.JAVA.id || AppCache.repoLanguage.id == Language.PYTHON.id || AppCache.repoLanguage.id == Language.JAVASCRIPT.id
```
khemrajrathore, line 91: alternatively we can also compare `AppCache.repoLanguage == Language.JAVA`

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
```diff
@@ -41,22 +41,63 @@ class HttpConnectionMetadataExporter(cpg: Cpg, ruleCache: RuleCache) {
...
+  def getLiteralsFromLanguageFiles: List[String] = {
+    val egressLiterals = cpg
+      .literal(URL_PATH_WITH_VARIABLE_SYMBOLS)
+      .filter(node => node.code.replaceAll(NON_ALPHANUMERIC, "") != "")
+      .filter(!_.code.matches(STRING_WITH_CONSECUTIVE_DOTS_OR_DOT_SLASH_OR_NEWLINE))
+      .inCall
+      .map(node =>
+        node.argument
+          .map(arg => {
+            if node.name == "require" then ""   // skip import statement
```
khemrajrathore, line 62: These nodes can be filtered out outside as 
```
inCall
.where(_.nameNot("require"))
.map(node => ....continue
```

Also I belive `import` also needs to be added along with `require`

src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
```diff
@@ -0,0 +1,93 @@
...
+      | BASE_HOST = "api/v1"
+      | API_4 =      BASE_HOST + "/ce/customers/{customerId}/scan/{repoId}"
+      | API_5 =      "api/v2" + "/ce/customers"
+      |""".stripMargin)
+
+  "Python code egresses" should {
+    "collect egress url for python code" in {
+      val propertyExporter          = new HttpConnectionMetadataExporter(cpg, ruleCache)
+      val egressesFromLanguageFiles = propertyExporter.getLiteralsFromLanguageFiles
+      print(egressesFromLanguageFiles)
```
khemrajrathore, line 79: Let remove the print statement

src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
```diff
@@ -0,0 +1,93 @@
...
+      egressesFromLanguageFiles.size shouldBe 6
+      egressesFromLanguageFiles shouldBe List(
+        "API_HOST/ce/something/customers/{customerId}/init",
+        "API_HOST/ce/customers/{customerId}/repo/{repoId}/file",
```
khemrajrathore, line 83: I see for python we don't replace the dynamic values `{customerId}` is present as is, but in Java the `${customerId}` is getting replaced this can be problematic for ingress matching. Would recommend a standard way to represent such values irrespective of language

src/test/scala/ai/privado/exporter/JavaLanguageEgressTest.scala
```diff
@@ -0,0 +1,91 @@
...
+      egressesFromLanguageFiles.size shouldBe 5
+      egressesFromLanguageFiles shouldBe List(
+        "api/v1/login",
+        "api/v1/user/meta",
+        "https://vineet%s/hellow",
+        "api/hellow",
+        "https://abc.com/planetary/apod?api_key=apiKey"
+      )
+    }
+  }
```
khemrajrathore, line 89: Can we add a test case for multiple assignments, I believe it will not give expected result. 
Ex - `"api/v1/" + "user/meta" + "/profile"`

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
nikvin15: Cool updated accordingly

src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
nikvin15: @khemrajrathore We are not doing anything extra to remove that for javascript. I think its because of underlying core logic of joern for python and javascript format functions. However I can have a look but this should not be a problem on matching as we remove all symbols when we match ingress and egress so do you think it is a blocker or I can spend time on this later ?

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
```diff
@@ -53,19 +53,19 @@ class HttpConnectionMetadataExporter(cpg: Cpg, ruleCache: RuleCache) {
   def getLiteralsFromLanguageFiles: List[String] = {
     val egressLiterals = cpg
       .literal(URL_PATH_WITH_VARIABLE_SYMBOLS)
-      .filter(node => node.code.replaceAll(NON_ALPHANUMERIC, "") != "")
+      .filter(node => ALPHABET.r.findFirstIn(node.code).isDefined)
```
khemrajrathore, line 56: Optionally we can also write this as `.where(_.code(ALPHABET))`
[Not a blocker]

khemrajrathore (approved):
LGTM

src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
khemrajrathore: If we take this precaution while matching, then no issue

src/main/scala/ai/privado/exporter/HttpConnectionMetadataExporter.scala
nikvin15: I will make it short in follow up PR for sure. Thanks for tip

nikvin15: @khemrajrathore It seems that to make it work for `_.code(<regex>)`, we have to write a regex that allows other characters as well, such as `(.*)[a-zA-Z](.*)`. Therefore, I believe findFirstIn would be better suited here.

PR #1002, opened 2024-03-08
LE-1096 | Fix ingressUrls append for method full name collection tagger

sriram-privado:
* Fix ingressUrls collection during method full name tagging.
* Make the ingressUrls into a list buffer.

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/collection/MethodFullNameCollectionTagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/collection/MethodFullNameCollectionTagger.scala
@@ -67,10 +67,14 @@ class MethodFullNameCollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends Col
     val methods           = scala.collection.mutable.HashMap.empty[Long, Method]
     val localMethodUrlMap = scala.collection.mutable.HashMap.empty[Long, String]
     for (methodCall <- methodCalls) {
-      val url                           = methodCall.argument.isLiteral.code.head
       var handlerMethod: Option[Method] = Option.empty[Method]
 
       breakable {
+        val url = methodCall.argument.isLiteral.code.headOption
+        if (url.isEmpty) {
+          break // continue loop, no URL is found as the first argument
+        }
+
         if (methodCall.argument.length < 2) { // we do not have enough arguments to get the handler method
           break                               // empty handler, continue loop
         }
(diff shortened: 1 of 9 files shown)


src/main/scala/ai/privado/languageEngine/java/tagger/collection/MethodFullNameCollectionTagger.scala
```diff
@@ -67,10 +67,14 @@ class MethodFullNameCollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends Col
     val methods           = scala.collection.mutable.HashMap.empty[Long, Method]
     val localMethodUrlMap = scala.collection.mutable.HashMap.empty[Long, String]
     for (methodCall <- methodCalls) {
-      val url                           = methodCall.argument.isLiteral.code.head
       var handlerMethod: Option[Method] = Option.empty[Method]
 
       breakable {
+        val url = methodCall.argument.isLiteral.code.headOption
+        if (url.isEmpty) {
+          break // continue loop, no URL is found as the first argument
```
pandurangpatil, line 75: If the logic is dependent on `URL` then `break` makes sense. If the logic is not dependent on the `URL` and it is just acting as metadata. Then we can handle it like `methodCall.argument.isLiteral.code.headOption.getOrElse("")` setting to empty. It might happen frontend failed to create the AST node in some cases, but because of that, we failed to identify the collection point. 

I am just thinking about the possibility.

sriram-privado: @pandurangpatil good observation. In this case, empty string or undefined both don't make any sense to move forward in the loop as url is mandatory for the tagging.

dbMundada: Not sure about ingress urls but logic Looks good to me.

pandurangpatil: @dbMundada in case we are not able to map the URL for collections. We show methdName on the dashboard for JS.

sriram-privado: @pandurangpatil in this case, the code for Kotlin or Java is of the format - 
```
get("/some-endpoint", handlerMethod)
```
So, having an empty or undefined URL is not a valid use-case, it has to be first argument literal.  So, I think its ok to ignore this here. 

Is there an example code you had in mind where we just show the methodname without URL ?

pandurangpatil: I don't have a precise sample for the use case. 
I guess we can proceed further for now.

pandurangpatil (approved):

PR #1007, opened 2024-03-11
fix(languageEgress): consider a string which has slash in the end

nikvin15:
# For Egress, consider a string which has slash in the end


hiteshbedre (approved):
LGTM

PR #1008, opened 2024-03-11
product level datamap: exclude false positive entries from egress

hiteshbedre:
Changes: Eliminating egress urls such as filename, aws urls or html template which is matching existing pattern(string having slash)

After adding this elimination pattern around 90% entries are reduced.

Also enabled one of test case which is marked ignored last time.


khemrajrathore (approved):
LGTM

PR #1009, opened 2024-03-12
API YamlLinkerPass support

ankit-privado:
- Added support to detect the internal flow having domain (i.e. `http://exampleService`) present inside the yaml file. 
- Made the implementation generic and can be used in anyLanguage, we only need to define the mapping rule between code and key.

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/go/passes/config/GoYamlLinkerPass.scala
+++ b/src/main/scala/ai/privado/languageEngine/go/passes/config/GoYamlLinkerPass.scala
@@ -1,15 +1,11 @@
-package ai.privado.languageEngine.java.passes.config
+package ai.privado.languageEngine.go.passes.config
 
-import io.shiftleft.codepropertygraph.generated.nodes.*
 import ai.privado.passes.YamlLinkerPass
 import io.shiftleft.codepropertygraph.generated.Cpg
 import io.shiftleft.codepropertygraph.generated.nodes.Literal
-import org.slf4j.LoggerFactory
 import io.shiftleft.semanticcpg.language.*
 
-class JavaYamlLinkerPass(cpg: Cpg) extends YamlLinkerPass(cpg) {
-  private val logger = LoggerFactory.getLogger(this.getClass)
-
+class GoYamlLinkerPass(cpg: Cpg) extends YamlLinkerPass(cpg) {
   override def matchingLiteralsToPropertyNode(propertyName: String): List[Literal] = {
     val propertyKey = propertyName.split("\\.").last
     cpg.literal
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
@@ -109,6 +109,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
   override def runOnPart(builder: DiffGraphBuilder, ruleInfo: RuleInfo): Unit = {
     val apiInternalSources = cpg.literal.code("(?:\"|')(" + ruleInfo.combinedRulePattern + ")(?:\"|')").l
     val propertySources    = cpg.property.filter(p => p.value matches (ruleInfo.combinedRulePattern)).usedAt.l
+    val serviceSource      = cpg.property.filter(p => p.value matches ".*(http|https):\\/\\/.*").usedAt.l
 
     // Support to use `identifier` in API's
     val identifierRegex = ruleCache.getSystemConfigByKey(Constants.apiIdentifier)
@@ -159,7 +160,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
         logger.debug("Using Enhanced API tagger to find API sinks")
         println(s"${Calendar.getInstance().getTime} - --API TAGGER V2 invoked...")
         sinkTagger(
-          apiInternalSources ++ propertySources ++ identifierSource,
+          apiInternalSources ++ propertySources ++ identifierSource ++ serviceSource,
           apis.methodFullName(commonHttpPackages).l ++ feignAPISinks ++ grpcSinks ++ soapSinks,
           builder,
           ruleInfo,
--- a/src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
@@ -30,7 +30,7 @@ import ai.privado.languageEngine.java.tagger.Utility.{GRPCTaggerUtility, SOAPTag
 import ai.privado.metric.MetricHandler
 import ai.privado.model.{Constants, Language, NodeType, RuleInfo}
 import ai.privado.tagger.PrivadoParallelCpgPass
-import ai.privado.tagger.utility.APITaggerUtility.sinkTagger
+import ai.privado.tagger.utility.APITaggerUtility.{SERVICE_URL_REGEX_PATTERN, sinkTagger}
 import ai.privado.utility.{ImportUtility, Utilities}
 import io.circe.Json
 import io.joern.dataflowengineoss.queryengine.{EngineConfig, EngineContext}
@@ -109,7 +109,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
   override def runOnPart(builder: DiffGraphBuilder, ruleInfo: RuleInfo): Unit = {
     val apiInternalSources = cpg.literal.code("(?:\"|')(" + ruleInfo.combinedRulePattern + ")(?:\"|')").l
     val propertySources    = cpg.property.filter(p => p.value matches (ruleInfo.combinedRulePattern)).usedAt.l
-    val serviceSource      = cpg.property.filter(p => p.value matches ".*(http|https):\\/\\/.*").usedAt.l
+    val serviceSource      = cpg.property.filter(p => p.value matches SERVICE_URL_REGEX_PATTERN).usedAt.l
 
     // Support to use `identifier` in API's
     val identifierRegex = ruleCache.getSystemConfigByKey(Constants.apiIdentifier)
@@ -140,7 +140,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
         logger.debug("Using brute API Tagger to find API sinks")
         println(s"${Calendar.getInstance().getTime} - --API TAGGER V1 invoked...")
         sinkTagger(
-          apiInternalSources ++ propertySources ++ identifierSource,
+          apiInternalSources ++ propertySources ++ identifierSource ++ serviceSource,
           apis,
           builder,
           ruleInfo,
@@ -149,7 +149,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
           privadoInputConfig.enableAPIDisplay
         )
         sinkTagger(
-          apiInternalSources ++ propertySources ++ identifierSource,
+          apiInternalSources ++ propertySources ++ identifierSource ++ serviceSource,
           feignAPISinks ++ grpcSinks ++ soapSinks,
           builder,
           ruleInfo,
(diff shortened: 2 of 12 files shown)


src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
```diff
@@ -109,6 +109,7 @@ class JavaAPITagger(cpg: Cpg, ruleCache: RuleCache, privadoInputConfig: PrivadoI
   override def runOnPart(builder: DiffGraphBuilder, ruleInfo: RuleInfo): Unit = {
     val apiInternalSources = cpg.literal.code("(?:\"|')(" + ruleInfo.combinedRulePattern + ")(?:\"|')").l
     val propertySources    = cpg.property.filter(p => p.value matches (ruleInfo.combinedRulePattern)).usedAt.l
+    val serviceSource      = cpg.property.filter(p => p.value matches ".*(http|https):\\/\\/.*").usedAt.l
```
HiteshMah-Jan, line 112: can we update the combinedRulePattern itself instead.

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
```diff
@@ -0,0 +1,40 @@
...
+      builder.addEdge(propertyNode, lit, EdgeTypes.IS_USED_AT)
+      builder.addEdge(lit, propertyNode, EdgeTypes.ORIGINAL_PROPERTY)
+    })
+  }
+
+  private def matchingLiteralsInGetEnvCalls(propertyName: String): List[Literal] = {
+    val propertyKey = propertyName.split("\\.").last
+    cpg.literal
+      .codeExact("\"" + propertyKey + "\"")
+      .filter(_.inCall.name("(?i).*getenv").nonEmpty)
```
HiteshMah-Jan, line 37: Please check how env is read in other languages and update the regex as per

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
```diff
@@ -0,0 +1,40 @@
+package ai.privado.passes
+
+import ai.privado.languageEngine.java.language.NodeStarters
+import ai.privado.tagger.PrivadoParallelCpgPass
+import io.shiftleft.codepropertygraph.generated.{Cpg, EdgeTypes}
+import io.shiftleft.codepropertygraph.generated.nodes.*
+import io.shiftleft.codepropertygraph.generated.nodes.JavaProperty
+import io.shiftleft.semanticcpg.language.*
+
+class YamlLinkerPass(cpg: Cpg) extends PrivadoParallelCpgPass[JavaProperty](cpg) {
```
kbatavia12, line 10: Why do we need a separate pass for this? We can do this inside the `JavaPropertyLinkerPass` itself.

src/main/scala/ai/privado/tagger/utility/APITaggerUtility.scala
```diff
@@ -71,18 +71,25 @@ object APITaggerUtility {
...
-        val apiNode     = flow.elements.last
+        val sourceNode = flow.elements.head
+        val apiNode    = flow.elements.last
         // Tag API's when we find a dataflow to them
         var newRuleIdToUse = ruleInfo.id
         if (ruleInfo.id.equals(Constants.internalAPIRuleId)) addRuleTags(builder, apiNode, ruleInfo, ruleCache)
         else {
-          val domain = getDomainFromString(literalCode)
+          val domain = {
+            if (sourceNode.originalPropertyValue.nonEmpty) {
```
HiteshMah-Jan, line 81: ? why is sourceNode.originalPropertyValue nonEmpty only for k8s domains ?

src/test/scala/ai/privado/languageEngine/java/passes/config/PropertiesFilePassTest.scala
```diff
@@ -300,6 +314,67 @@ class XMLPropertyTests extends PropertiesFilePassTestBase(".xml") {
...
+        |    HttpClient httpClient = HttpClients.createDefault();
+        |    HttpGet getRequest = new HttpGet(apiUrl);
+        |    HttpResponse response = httpClient.execute(getRequest);
+        |  }
+        |}
+        |""".stripMargin
+
+  override val propertyFileContents: String = ""
+  "fetch api" should {
+    "ooo" in {
```
kbatavia12, line 345: Can we use a descriptive test name please?

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  private def connectProperty(propertyNode: JavaProperty, builder: DiffGraphBuilder): Unit = {
+    matchingLiteralsInGetEnvCalls(propertyNode.name).foreach(lit => {
+      builder.addEdge(propertyNode, lit, EdgeTypes.IS_USED_AT)
+      builder.addEdge(lit, propertyNode, EdgeTypes.ORIGINAL_PROPERTY)
+    })
+  }
+
+  private def matchingLiteralsInGetEnvCalls(propertyName: String): List[Literal] = {
```
kbatavia12, line 33: If we want to ensure that there is good coverage, we should look at other ways through which YAML config is consumed in code. @HiteshMah-Jan can confirm this.

kbatavia12 (changes requested):

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
ankit-privado: Yes, but I believe javaPropertyLinkerPass is intended for Java's .properties files, which handle a specific use case where we fetch the value using the getProperty call. Yes, we can modify the code to use JavaPropertyLinkerPass, but handling the YAML related functionality in a separate pass would make it loosely coupled.

src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
ankit-privado: yes sure, I need to raise the PR for rule change also.

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
ankit-privado: yes i need to add support for other language too. will update the PR.

src/main/scala/ai/privado/tagger/utility/APITaggerUtility.scala
ankit-privado: this is to check if the domain is present inside `YAML` file or not.

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
kbatavia12: The intended purpose of the `JavaPropertyLinkerPass` is to link properties to their respective use sites. We'll have many passes if we want to do it for other property files as well.

kbatavia12: You can simply add an `if` condition for YAML properties in the `JavaPropertyLinkerPass`.

ankit-privado: yes agree, In that case we will need to make the `JavaPropertyLinkerPass` more generic, which will require refactoring. @HiteshMah-Jan should we go with one pass for all `property` related linking?

ankit-privado: As discussed with @HiteshMah-Jan, we are going with separate pass.

src/main/scala/ai/privado/languageEngine/java/tagger/sink/JavaAPITagger.scala
ankit-privado: hard coded in code itself as of now. 
CC: @HiteshMah-Jan

src/main/scala/ai/privado/passes/YamlLinkerPass.scala
ankit-privado: we will add support for other language in future PRs.

src/main/scala/ai/privado/tagger/utility/APITaggerUtility.scala
ankit-privado: handled the domain resolution for k8s.

src/main/scala/ai/privado/languageEngine/go/passes/config/GoYamlLinkerPass.scala
```diff
@@ -0,0 +1,16 @@
...
+import io.shiftleft.codepropertygraph.generated.nodes.Literal
+import io.shiftleft.semanticcpg.language.*
+
+class GoYamlLinkerPass(cpg: Cpg) extends YamlLinkerPass(cpg) {
+  override def matchingLiteralsToPropertyNode(propertyName: String): List[Literal] = {
+    val propertyKey = propertyName.split("\\.").last
+    cpg.literal
+      .codeExact("\"" + propertyKey + "\"")
+      .filter(_.inCall.name("(?i).*getenv").nonEmpty)
+      .toList
```
khemrajrathore, line 14: I believe this same kind of thing we are already doing in the Language specific PropertyLinkerPass, didn't understood the usage of this.

ankit-privado: Actually `javaPropertyLinkerPass` is intended for Java's `.properties` files, which handle a specific use case where we fetch the env value using the `getProperty` call. Yes, we can do this in JavaPropertyLinkerPass also, but handling the YAML related functionality in a separate pass would make it loosely coupled. I had the same discussion with @HiteshMah-Jan regarding this, he also suggested to make separate pass for it.

khemrajrathore: `javaPropertyLinkerPass is intended for Java's .properties files` This statement doesn't hold true, the linking is done irrespective of property being from `.properties` or not

khemrajrathore: I would also recommend seperate pass, but the logic to create a new pass should be on the way the property is linked. Ex - a pass which links `property to the env calls`, instead of a separate linking pass for JSON, YAML

src/test/scala/ai/privado/languageEngine/java/passes/config/JavaYamlLinkerPassTest.scala
```diff
@@ -0,0 +1,199 @@
...
+      |import org.apache.http.HttpResponse;
+      |import org.apache.http.client.HttpClient;
+      |import org.apache.http.client.methods.HttpGet;
+      |import org.apache.http.impl.client.HttpClients;
+      |
+      |public class APICaller {
+      |  private String apiUrl;
+      |
+      |  public makeCall() {
+      |    apiUrl = System.getenv("API_URL");
```
khemrajrathore, line 51: This is used to get Environment variables, are we saying this operation happens only with Yaml file in Java?

src/main/scala/ai/privado/languageEngine/go/passes/config/GoYamlLinkerPass.scala
ankit-privado: > `javaPropertyLinkerPass is intended for Java's .properties files` This statement doesn't hold true, the linking is done irrespective of property being from `.properties` or not

actually its written in the code that `javapropertyLinkerPass` is for `.properties` file here but yes agree what you said, we should have one pass for all similar kind for fetching mechanism irrespective of property file extension. 

I think we also need to do some refactoring, the way current code written is specific to file extension.

karan-batavia: The comment says so @ankit-privado. The comment is written by David AFAIK, he might not have had the full context of what we were trying to do with the pass. The current pass takes all the properties, and tries to link them. As for the scope of this PR, you can either write your logic inside the JavaPropertyLinker, or do a refactor as @khemrajrathore suggested.

ankit-privado: yes @karan-batavia, I had a discussion with @khemrajrathore, we will take similar kind of approach to solve this use case in future but as of now we are moving with file extension approach.

khemrajrathore (approved):
LGTM

PR #1010, opened 2024-03-12
Store Egress from Code in a different key and remove parentheses from variable string

nikvin15:
### Egress From Code Changes
- We need to analysis Egress From Code so moving them in a separate key 
- `<operator>.formatString` itself has some call nodes in case of python so instead of doing for language specific, replacing `{}` with empty string 
- Test cases were failing because of known language so explicitly set `AppCache.repoLanguage` in test cases

### DIFF
--- a/src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
+++ b/src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
@@ -23,12 +23,12 @@
 
 package ai.privado.exporter
 
-import ai.privado.cache.RuleCache
+import ai.privado.cache.{AppCache, RuleCache}
 import ai.privado.entrypoint.PrivadoInput
 import ai.privado.exporter.HttpConnectionMetadataExporter
 import ai.privado.languageEngine.javascript.JavascriptTaggingTestBase
 import ai.privado.languageEngine.python.{PrivadoPySrc2CpgFixture, PrivadoPySrcTestCpg}
-import ai.privado.model.ConfigAndRules
+import ai.privado.model.{ConfigAndRules, Language}
 import ai.privado.tagger.sink.RegularSinkTagger
 import org.scalatest.BeforeAndAfterAll
 import org.scalatest.matchers.should.Matchers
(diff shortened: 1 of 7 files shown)


hiteshbedre (approved):
LGTM

src/test/scala/ai/privado/exporter/PythonLanguageEgressTest.scala
```diff
@@ -74,17 +74,22 @@ class PythonLanguageEgressTest extends PrivadoPySrc2CpgFixture {
       | API_5 =      "api/v2" + "/ce/customers"
       |""".stripMargin)
 
+  override def beforeAll(): Unit = {
+    AppCache.repoLanguage = Language.PYTHON
```
khemrajrathore, line 78: This can result in test case flakiness across, this global object needs to be converted to an instance object instead
cc- @nikvin15 @hiteshbedre

nikvin15: I took a reference from this file https://github.com/Privado-Inc/privado-core/blob/main/src/test/scala/ai/privado/languageEngine/java/JavaTaggingTestBase.scala#L56. I don't have much idea but is this also not same 🤔 I thought we reset the repoLanguage for every test case.

khemrajrathore: The reference you took is problematic too and also needs to be changed. Usage of global objects often results in overwriting those values in test cases.

nikvin15: I see in javascript every suite runs on isolated environment so `AppCache.repoLanguage` will reset for each suite. If that is not the case and I understand you correctly. Could you guide me how to override `AppCache.repoLanguage ` because this is being used directly in the code so I didn't get the instance clue.

nikvin15: Synced offline with @khemrajrathore this is a backlog item for privado core so ignoring for now

PR #1013, opened 2024-03-13
Ruby/database schema

khemrajrathore:
The scope of the PR is to read `config/mongoid.yaml` file in ruby and retrieve the database client info, using that map it to ruby class which corresponds to all the collections/tables to that database

Ones we get the mapping of all the collections in a database, store that info in DatabaseDetailsCache, and export the schema


src/main/scala/ai/privado/languageEngine/ruby/tagger/schema/RubyMongoSchemaMapper.scala
```diff
@@ -0,0 +1,209 @@
...
+
+      Here we are interested in property nodes with values `default_production` and `noncustomer_production`
+     */
+    val clientNodes =
+      cpg.property.where(_.file.name(".*config/mongoid[.](yml|yaml)")).where(_.name(".*clients.*database")).l
+
+    val deploymentTypeSet = mutable.HashSet[String]()
+    val clientsSet        = mutable.HashSet[String]()
+
+    clientNodes.foreach(client => {
```
karan-batavia, line 52: Instead of using a foreach here, can we not use a `ParalleCpgPass`?

karan-batavia (approved):
LG, just one comment

src/main/scala/ai/privado/languageEngine/ruby/tagger/schema/RubyMongoSchemaMapper.scala
khemrajrathore: The reason I remember is, there is dependency of clientNodes to build `deploymentTypeSet`, `clientsSet` and few other operations, like clientNodes is again used on line number 94 to generate the db information and Database details cache

khemrajrathore: There is possibility but that will require the pass to be divided into more granular passes and refactoring. Which we can focus on in near future

PR #1016, opened 2024-03-13
[SQLParser] Fix SQL lineNumber when query is found in .sql file

karan-batavia:
This PR includes a fix to display correct `lineNumber` of columns in a SQL query. This only includes queries from `.sql` files, as those queries are parsed individually.

### DIFF
--- a/src/main/scala/ai/privado/utility/SQLParser.scala
+++ b/src/main/scala/ai/privado/utility/SQLParser.scala
@@ -116,7 +116,7 @@ object SQLParser {
           val sqlTable = createSQLTableItem(createStmt.getTable)
           val columnList = columns.map(columnName => {
             val lineColumn = getLineAndColumnNumber(sqlQuery, columnName)
-            SQLColumn(columnName, lineColumn._1 + sqlTable.lineNumber, lineColumn._2)
+            SQLColumn(columnName, lineColumn._1, lineColumn._2)
           })
           Some(List(SQLQuery(SQLQueryType.CREATE, sqlTable, columnList)))
         case _ =>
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/privado/utility/SQLParser.scala
```diff
@@ -205,10 +206,16 @@ object SQLNodeBuilder {
     builder.addEdge(tableNode, fileNode, EdgeTypes.SOURCE_FILE)
 
     queryModel.column.zipWithIndex.foreach { case (queryColumn: SQLColumn, columnIndex) =>
+      /* As queries from .sql files are processed individually,
+        an offset equal to the lineNumber of the query in the original file - 1 is added to the isolated column lineNumber */
+      val lineNumber = fileName match
+        case Some(f) if f.endsWith(".sql") => queryColumn.lineNumber + queryLineNumber - 1
+        case _                             => queryColumn.lineNumber
```
khemrajrathore, line 213: I believe this will be needed across and not specifically for .sql files? Add a test case for this case as well

karan-batavia: Nope, just for .sql files. Other files ( ex -> schema.rb ) files give correct line numbers. They have their individual tests as well.

khemrajrathore: schema.rb is parsed by ruby parser and not by this utility function

karan-batavia: Agreed, but all the columns and tables are finally passed to this `SQLNodeBuilder`.

khemrajrathore (approved):
LGTM

PR #1019, opened 2024-03-18
add support to tag api by matching url like parameter

khemrajrathore:
This PR contains
- Support for tagging API by methodFullName by specifying the regex in `privado->config->systemConfig->file.yaml` under the key `apiMethodFullNames`
- Support for tagging API by deriving API client by looking at the parameter names, this has been kept under the flag `enable-api-by-parameter` for now

This PR also proposes to refactor the current API tagger implementation such as tagging API sinks as `API_SINK_MARKED`, based on respective detection logic.
All these tagged API's can be collected in the APITagger and dataflow from literal to them can be computed and tagged appropriately


hiteshbedre (approved):
LGTM

PR #1021, opened 2024-03-19
Property Pass Restructuring 

ankit-privado:
This PR involves restructuring the PropertyLinkerPass for all languages. Below are the changes included in this PR:

1. An abstract class named EnvPropertyLinkerPassBase has been introduced, which will be implemented by language-specific LinkerPasses. It includes common utility methods and property linker logic.
2. The propertyLinkerPass has been renamed to EnvPropertyLinkerPass because we are handling the use case of linking literal nodes to environment properties.
3. The yaml linker pass has been removed and merged with specific linker passes.
4. Support for RubyEnvPropertyLinkerPass and its unit test has been added.
5. AnnotationLinker is separated out from EnvLinker for Java. 
6. DBLinker is separated for Python.


src/main/scala/ai/privado/passes/PropertyEnvLinkerPassBase.scala
```diff
@@ -0,0 +1,34 @@
...
+import overflowdb.{BatchedUpdate, Node}
+
+/** This pass creates a graph layer for env call from property node
+  */
+
+abstract class PropertyEnvLinkerPassBase(cpg: Cpg) extends PrivadoParallelCpgPass[JavaProperty](cpg) {
+
+  val logger = LoggerFactory.getLogger(getClass)
+  override def generateParts(): Array[_ <: AnyRef] = {
+    cpg.property.iterator
```
khemrajrathore, line 18: As discussed, this is incorrect, as in `generateParts` the call where we intend to link Ex - `getEnv` calls `annotations` needs to be added, because these are limited in number as compared to propertyNodes, which can easily shoot up

ankit-privado: yes right, I was testing something, i will make the change and once the PR is ready for review, i will notify you.

ankit-privado: Yes done.

PR #1032, opened 2024-03-26
csharp: add collections support

tuxology:
* Adds basic collections support for AspNetCore MVC style routing
 * Supports Attribute Routing and not Conventional Routing
 * Ref: https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/routing

### DIFF
--- a/src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
+++ b/src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
@@ -38,6 +38,8 @@ class CollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelCp
   protected val methodUrlMap: mutable.HashMap[Long, String] = mutable.HashMap[Long, String]()
   protected val classUrlMap: mutable.HashMap[Long, String]  = mutable.HashMap[Long, String]()
   private val logger                                        = LoggerFactory.getLogger(this.getClass)
+  enum NodeType:
+    case Method, Class
 
   override def generateParts(): Array[RuleInfo] =
     ruleCache.getRule.collections.filter(_.catLevelTwo == Constants.annotations).toArray
@@ -75,12 +77,12 @@ class CollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelCp
     )
   }
 
-  private def getUrlFromAnnotationsCode(annotation: Annotation): String = {
+  private def getUrlFromAnnotationsCode(annotation: Annotation, nodeType: NodeType): String = {
     // If we have [controller] in annotation code, then replace with the class name of controller
     //
-    //    [Route("api/some/[controller]")]                      <-- /api/some/email
+    //    [Route("api/some/[controller]")]                      <-- /api/some/email            [CASE A]
     //    class EmailController {
-    //      [HttpGet("[controller]/[action]/{id}")]             <-- /api/some/email/edit/{id}
+    //      [HttpGet("[controller]/[action]/{id}")]             <-- /api/some/email/edit/{id}  [CASE B]
     //      public IActionResult Edit(int id)
     //      {
     //        return ControllerContext.MyDisplayRouteInfo(id);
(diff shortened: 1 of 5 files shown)


src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,139 @@
...
+    //      {
+    //        return ControllerContext.MyDisplayRouteInfo(id);
+    //      }
+    //    }
+    //
+    val controllerName = if annotation.code.contains("[controller]") then {
+      cpg.typeDecl
+        .where(_.annotation.codeExact(annotation.code))
+        .where(_.name(".*Controller"))
+        .name
```
khemrajrathore, line 94: As we already have the `annotation` node, we can directly go to the respective typeDecl by `annotation.typeDecl`, instead `cpg.typeDecl`

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,139 @@
...
+        .headOption
+        .getOrElse("")
+        .split("Controller")
```
khemrajrathore, line 97: We can also trim `Controller` to get to the class name, assuming it is `ClassNameController`, as triming is a safer operation than split

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,139 @@
...
+        .getOrElse("")
+        .toLowerCase
+    } else {
+      ""
+    }
+
+    // Now we extract [action] seen in snippet above. Only method support for now since we encounter them most here
+    // TODO: Support [action] replacement for classes as well eg Route("[controller]/[action]")
+    val actionName = if annotation.code.contains("[action]") then {
+      cpg.method.where(_.annotation.codeExact(annotation.code)).name.headOption.getOrElse("").toLowerCase
```
khemrajrathore, line 108: similarly we can do `annotation.method`

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+      ""
+    }
+    annotation.code
+      .replaceAll(".*\"(.*?)\".*", "/$1")
```
khemrajrathore, line 113: Here are we looking to pick the content present inside the `""` double quotes?
Please add a comment explaining the purpose of this

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,139 @@
...
+      builder,
+      collectionMethodsCache,
+      ruleInfo,
+      ruleCache,
+      methodUrlMap = methodUrlMap,
+      classUrlMap = classUrlMap
+    )
+  }
+
+  private def getUrlFromAnnotationsCode(annotation: Annotation): String = {
```
khemrajrathore, line 78: When calling this function, we know before hand if this annotation is placed on a method or typeDecl, that information can be passed to this function and that node can be used for below queries, by this quering on cpg can be limited

src/test/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTaggerTest.scala
```diff
@@ -0,0 +1,52 @@
...
+              |    }
+              |  }
+              |}
+              |""".stripMargin,
+            "Test.cs"
+          )
+        )
+      )
+
+      val List(emailMethod) = cpg.method.nameExact("GetEmail").l
```
khemrajrathore, line 42: Can we cover more test cases, Ex- different ways to pass the routes in annotation, what if route present on class not on method etc

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
tuxology: Isn't `trim` just remoiing empty spaces? We can `stripSuffix` though - that might be safer

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
tuxology: We do need both queries to run on types as well as methods given the different ways in which routes can be defined, but I think I got another idea from this. Will implement this 👍

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,210 @@
...
+    //      {
+    //        return ControllerContext.MyDisplayRouteInfo(id);
+    //      }
+    //    }
+    //
+    val controllerName = if annotation.code.contains("[controller]") then {
+      nodeType match
+        case NodeType.Class =>
+          annotation.start.typeDecl
+            .where(_.name(".*Controller"))
```
khemrajrathore, line 100: Do we need to have this check, this can prevent us from getting results for controller which don't follow this convention?

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
```diff
@@ -0,0 +1,210 @@
...
+            .matches(".*\\[action\\].*") && !annotation.code.matches(".*\\[action\\].*")
+        then {
+          if annotation.code.matches(".*\"(.*?)\".*") then {
+            s"$actionName" + annotation.code.replaceAll(
+              ".*\"(.*?)\".*",
+              "/$1"
+            ) // CASE D: we append action name for case like [HttpGet] where there is no argument
+          } else if !annotation.code.matches(".*\"(.*?)\".*") then {
+            s"$actionName" // CASE Ca: No brackets in annotation, means we don't need to append
+          } else {
```
khemrajrathore, line 165: This looks unreachable, above we can have `if` and `else`, no need for `else if`

khemrajrathore:
@tuxology Solid coverage in test case💪

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
karan-batavia: Controllers are always named like this in C#.

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
tuxology: Added `else if` to explicitly check for the case where we have and don't have `".*\"(.*?)\".*"` We might have other cases we will eventually have to support - in such cases, it would be nice just just send empty string for now instead of wrong `actionName` which might confuse us later.

src/main/scala/ai/privado/languageEngine/csharp/tagger/collection/CollectionTagger.scala
tuxology: This is a requirement in ASP.NET Core as @karan-batavia said. The first part of the Class name is then used to replace `[controller]` string in attribute.

karan-batavia (approved):

PR #1035, opened 2024-03-27
fix: incorrect PHP parser vendoring

pawan-privado:
Vendored utilities are copied to a different path as part of Docker's build steps (ref. to the last couple of `COPY` commands in `Dockerfile`). Because this was not accounted for in the previous integration PR, there was an issue running a scan on PHP repositories. This PR fixes it.


khemrajrathore (approved):
LGTM

PR #1044, opened 2024-04-01
Repo Name output fix

ankit-privado:
fixed the repoName string in AppCache.

### DIFF
--- a/src/main/scala/ai/privado/cache/AppCache.scala
+++ b/src/main/scala/ai/privado/cache/AppCache.scala
@@ -46,6 +46,6 @@ class AppCache(
   def init(path: String): Unit = {
     this.scanPath = path.stripSuffix("/")      // Scan Path of the repo on the host machine
     this.localScanPath = getRepoScanPath(path) // scan path perceived by the binary (can be different inside docker)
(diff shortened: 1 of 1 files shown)


src/main/scala/ai/privado/cache/AppCache.scala
```diff
@@ -46,6 +46,6 @@ class AppCache(
   def init(path: String): Unit = {
     this.scanPath = path.stripSuffix("/")      // Scan Path of the repo on the host machine
     this.localScanPath = getRepoScanPath(path) // scan path perceived by the binary (can be different inside docker)
-    this.repoName = this.localScanPath.split("/\\\\").last
+    this.repoName = this.localScanPath.split("[/\\\\]").last
```
karan-batavia, line 49: Can we use `lastOption` instead of `last`? Can result in a runtime error.

karan-batavia (changes requested):

src/main/scala/ai/privado/cache/AppCache.scala
ankit-privado: done.

pandurangpatil (approved):

karan-batavia (approved):

PR #1048, opened 2024-04-03
IndentifierTagger split into multiple taggers

pandurangpatil:
Made changes to split the IdentifierTagger into multiple chnuks to divide the smaller chunks of the task more evenly to improve the CPU utilisation


src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
```diff
@@ -0,0 +1,325 @@
...
+  val cachedFieldAccess = cpg.method
+    .fullNameExact(Operators.fieldAccess, Operators.indirectFieldAccess)
+    .callIn
+    .dedup
+    .l
+  val derivedSourceTaggerParts = ruleCache.getRule.sources
+    .flatMap(rule => {
+      cachedMember
+        .name(rule.combinedRulePattern)
+        .map(memberNode => DerivedSourceTaggerPart(memberNode.typeDecl, memberNode, rule))
```
khemrajrathore, line 278: There can be a further optimization, as after DirectSourceTagging, we know the member and the ruleId against which it was tagged

src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
```diff
@@ -0,0 +1,325 @@
...
+        .l
+    })
+    .toArray
+  val secondLevelDerivedSourceTaggerParts = ruleCache.getRule.sources
+    .flatMap(rule => {
+      cachedMember
+        .name(rule.combinedRulePattern)
+        .map(_.typeDecl)
+        .distinctBy(_.fullName)
+        .map(memberTypeDecl => SecondLevelDerivedSourceTaggerPart(memberTypeDecl, rule))
```
khemrajrathore, line 288: Similar optimization as above can be achieved via tagging

src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
```diff
@@ -0,0 +1,325 @@
...
+      })
+
+    // To Mark all field Access and getters
+    cpgNodeCache.tagAllFieldAccessAndGetters(builder, typeDeclVal, taggerPart.ruleInfo, typeDeclMemberName, true)
+  }
+}
+
+case class ExtendingDerivedSourceTaggerPart(typeDecl: TypeDecl, memberType: SecondLevelDerivedSourceTaggerPart)
+case class SecondLevelDerivedSourceTaggerPart(typeDecl: TypeDecl, ruleInfo: RuleInfo)
+case class DerivedSourceTaggerPart(typeDecl: TypeDecl, member: Member, ruleInfo: RuleInfo)
```
khemrajrathore, line 258: Storing `ruleInfo` against each node can be a memory intensive operation, we ideally only need the `ruleId`

pandurangpatil: No, it doesn't matter. Mapping RuleInfo vs ruleId will have a similar memory footprint, as it will only hold the reference to RuleInfo (we are not cloning the RuleInfo). The reference will occupy same size as that of Int.

src/main/scala/ai/privado/languageEngine/java/tagger/PrivadoTagger.scala
```diff
@@ -74,8 +74,11 @@ class PrivadoTagger(cpg: Cpg) extends PrivadoBaseTagger {
 
     new SqlQueryTagger(cpg, ruleCache).createAndApply()
 
-    new IdentifierTagger(cpg, ruleCache, taggerCache).createAndApply()
-
+    val nodeCache = CPGNodeCacheForSourceTagger(cpg, ruleCache)
+    new DirectNodeSourceTagger(cpg, nodeCache, ruleCache, taggerCache).createAndApply()
+    new FirstLevelDerivedSourceTagger(cpg, nodeCache, ruleCache, taggerCache).createAndApply()
+    new OCDDerivedSourceTagger(cpg, nodeCache, ruleCache, taggerCache).createAndApply()
+    new ExtendingDerivedSourceTagger(cpg, nodeCache, ruleCache, taggerCache).createAndApply()
```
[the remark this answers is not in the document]
pandurangpatil: Good suggestion.

src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
pandurangpatil: It won't make much difference except the memory footprint. I was thinking about the approach you were suggesting if it would have increased the memory consumption by a large margin. As we are holding these caches in memory. 
However, after going through a test run, I didn't find memory consumption to be very high. 

refer this screenshot of CPU and memory usage - https://drive.google.com/file/d/1T5fupM51-t1IJqg16VcZLCYwU8T_GFXD/view?usp=sharing - (Usage before 1PM is for tagger steps. Post 1 PM starts with API Tagger and reachableByFlows.)
You can also refer to this granular matrix - https://docs.google.com/spreadsheets/d/1H7XOeRRwLRm-5CL74jYyo08CfygzgY8LUYwSfvqKv9o/edit?usp=sharing

src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
pandurangpatil: refer to my above comment

src/main/scala/ai/privado/languageEngine/java/tagger/PrivadoTagger.scala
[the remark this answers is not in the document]
pandurangpatil: done

src/main/scala/ai/privado/languageEngine/java/tagger/source/SourceTagger.scala
```diff
@@ -0,0 +1,325 @@
...
+        taggerPart.ruleInfo.id
+      )
+    })
+
+    // To Mark all field Access and getters
+    cpgNodeCache.tagAllFieldAccessAndGetters(builder, typeDeclVal, taggerPart.ruleInfo, typeDeclMember.name, true)
+  }
+}
+
+/** Step 2.1 =>
```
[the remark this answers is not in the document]
pandurangpatil: done

src/test/scala/ai/privado/languageEngine/java/DummyJavaTest.scala
```diff
@@ -1,33 +1,36 @@
 package ai.privado.languageEngine.java
 
 import ai.privado.model.Constants
+import ai.privado.testfixtures.JavaFrontendTestSuite
 import io.shiftleft.semanticcpg.language.*
 import io.circe.Json
 import io.circe.syntax.EncoderOps
 
-class DummyJavaTest extends JavaTestCpgBase {
+class DummyJavaTest extends JavaFrontendTestSuite {
```
khemrajrathore, line 9: Questions
- How to use `custom rule` for a test case?
- Similarly how to use `PrivadoInput`?
- If we intend to write any test case on dataFlowCache, auditCache, s3DatabaseDetailsCache, appCache propertyFilterCache How we can achieve that?

The reason behind not following Joern's style of writing source code directly is
- In privado-core the content of sample source code used is more than that of joern and spanning across multiple files, leading to the source code occupying a lot of the test case space
- If we use a directory, similar code, which we often end up repeating, can be placed at one


Regarding 
```
Also in earlier case for each
`in` clause will execute the scan every time for sub sections of the
unit test. With the updated implementation, now the scan will be invoked
only once for each `should` section.
```

The 3 different styles of invoking the frontend `withCpg`, `withJson`, `withCpgAndJson` are proposed to fit the different needs of writing a test case.
Example - sometimes we would write test case only on the cpg, sometimes only on outputJson, and sometimes will need both.

The invocation inside `in` was done to show the different usages, the same can be achieved by using `withCpgAndOutputJson`, it that seems inefficient, we can introduce a similar `generateScanResult` on top of `CPG` to achieve what has been proposed

src/test/scala/ai/privado/testfixtures/PrivadoBaseTestFixture.scala
```diff
@@ -0,0 +1,36 @@
...
+    with Matchers
+    with BeforeAndAfterAll
+    with BeforeAndAfterEach
+    with Inside {
+
+  private val cpgs = mutable.ArrayBuffer.empty[TestCpg]
+
+  def code(code: String): T = {
+    val newCpg = testCpgFactory().moreCode(code)
+    cpgs.append(newCpg)
```
khemrajrathore, line 23: Can this lead to holding a lot of memory?

src/test/scala/ai/privado/testfixtures/TestCpg.scala
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  private def checkGraphEmpty(): Unit = {
+    if (_graph.isDefined) {
+      throw new RuntimeException("Modifying test data is not allowed after accessing graph.")
+    }
+  }
+
+  override def close(): Unit = {
+    _graph.foreach(_.close())
```
khemrajrathore, line 88: As it is a test case, whose cpg, won't be loaded again. Any pros/cons of this step?

src/test/scala/ai/privado/testfixtures/PrivadoBaseTestFixture.scala
pandurangpatil: Unit tests shouldn't code samples shouldn't end up creating too much of memory footprint.

khemrajrathore: As these will be references, should be ok

src/test/scala/ai/privado/testfixtures/TestCpg.scala
pandurangpatil: That's the default implementation of close. As TestCpg is holding its own `_graph` it should be closed as Cpg implements AutoClosable, it should do the clean up of its resources.

src/test/scala/ai/privado/languageEngine/java/DummyJavaTest.scala
pandurangpatil: > Questions
> 
> * How to use `custom rule` for a test case?
> * Similarly how to use `PrivadoInput`?
> * If we intend to write any test case on dataFlowCache, auditCache, s3DatabaseDetailsCache, appCache propertyFilterCache How we can achieve that?

```
    val privadoInput = PrivadoInput()
    val cpg = code("""..
        |""".stripMargin).withPrivadoInput(privadoInput).generateScanResult()
```
there are similar `withRuleCache`, `withAuditCache` etc implementations to pass on respective input while creating the CPG in the unit test itself.  

> The reason behind not following Joern's style of writing source code directly is
> 
> * In privado-core the content of sample source code used is more than that of joern and spanning across multiple files, leading to the source code occupying a lot of the test case space

With sample code in other places, unit tests no longer remain readable. I think even if the unit test code becomes more in a file that should be acceptable as long as it remains more readble. A probable way out to reduce the size of the file is to limit one combination of samples per unit test file.

> * If we use a directory, similar code, which we often end up repeating, can be placed at one

If there are more combinations of unit tests to be written for the same code sample combination. One should write those additional unit tests inside the same `should` section with more `in` sections. 

> Regarding
> 
> ```
> Also in earlier case for each
> `in` clause will execute the scan every time for sub sections of the
> unit test. With the updated implementation, now the scan will be invoked
> only once for each `should` section.
> ```
> 
> The 3 different styles of invoking the frontend `withCpg`, `withJson`, `withCpgAndJson` are proposed to fit the different needs of writing a test case. Example - sometimes we would write test case only on the cpg, sometimes only on outputJson, and sometimes will need both.
> 
> The invocation inside `in` was done to show the different usages, the same can be achieved by using `withCpgAndJson`, it that seems inefficient, we can introduce a similar `generateScanResult` on top of `CPG` to achieve what has been proposed

Yea that's perfectly understood, but devs will follow the way given in the sample code. They might not think over the optimisations while implementing.

src/test/scala/ai/privado/languageEngine/java/tagger/source/JavaSourceTaggingTest.scala
```diff
@@ -0,0 +1,117 @@
...
+        |       System.out.println(user);
+        |   }
+        |}
+        |""".stripMargin)
+      .generateScanResult()
+
+    "tag a userone and user identifier" in {
+      val identifierNodes = cpg.identifier.where(_.tag.nameExact("catLevelOne").valueExact("DerivedSources")).l
+      identifierNodes.size shouldBe 5
+      identifierNodes.code.l shouldBe List("userone", "userone", "userone", "user", "user")
```
khemrajrathore, line 114: One suggestion while writing test case is to also include some metadata like linenumber, which can help in relating in  what all it getting tag, as currently just by looking at 5 identifiers and there code, is not self explainatory

khemrajrathore (dismissed):
LGTM

src/test/scala/ai/privado/languageEngine/java/tagger/source/JavaSourceTaggingTest.scala
pandurangpatil: make sense

pandurangpatil: done

khemrajrathore (approved):
LGTM

PR #1059, opened 2024-04-04
Wrap PHP entity tagging using `Try`

pawan-privado:
The long-term and best solution is to handle AST-related anomalies either when constructing the AST or by introducing an additional pass. This PR is more of a short-term solution as discussed within the team.


src/main/scala/ai/privado/languageEngine/php/tagger/source/IdentifierTagger.scala
```diff
@@ -66,7 +70,12 @@ class IdentifierTagger(cpg: Cpg, ruleCache: RuleCache, taggerCache: TaggerCache)
       addRuleTags(builder, member, ruleInfo, ruleCache)
     })
 
-    tagObjectOfTypeDeclHavingMemberName(builder, rulePattern, ruleInfo)
+    Try(tagObjectOfTypeDeclHavingMemberName(builder, rulePattern, ruleInfo)) match {
+      case Success(_) =>
+      case Failure(reason) =>
+        logger.error("Error while tagging entities")
+        logger.error(reason.toString)
+    }
```
khemrajrathore, line 78: add a debug log as well, with rule details for future handling and replication

khemrajrathore (approved):
LGTM

PR #1060, opened 2024-04-08
IndentifierTagger split into multiple taggers (#1048)

pandurangpatil:
* IndentifierTagger split into multiple taggers

Made changes to split the IdentifierTagger which acts as source tagger into multiple chnuks to divide the smaller chunks of the task more evenly to improve the CPU utilisation

* review comment fixes

* Refactored base Test fixtures

Refactored base implementation of test fixtures and made it more simpler to write the code in test case itself. Also in earlier case for each `in` clause will execute the scan every time for sub sections of the unit test. With the updated implementation, now the scan will be invoked only once for each `should` section.

* Java source tagging unit tests

* Comparison report diff fixes

* Revert of earlier deleted file

Reverted earlier deleted file and kept both the approaches

* review comment fixes


khemrajrathore (approved):
LGTM

PR #1064, opened 2024-04-11
use possibleTypes in identifierTagger of JS

khemrajrathore:
We need to do a deep analysis into when to use `possibleTypes` and `typeFullName`, `methodFullName`, `dynamicTypeHintFullName` etc

We may also need to make changes in audit report or any other place if we see a gap

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ ThisBuild / version      := sys.env.getOrElse("BUILD_VERSION", "dev-SNAPSHOT")
 // parsed by project/Versions.scala, updated by updateDependencies.sh
 
 val cpgVersion        = "1.6.10"
-val joernVersion      = "2.0.323"
+val joernVersion      = "2.0.324"
 val overflowdbVersion = "1.190"
 val requests          = "0.8.0"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -8,7 +8,7 @@ ThisBuild / version      := sys.env.getOrElse("BUILD_VERSION", "dev-SNAPSHOT")
 // parsed by project/Versions.scala, updated by updateDependencies.sh
 
 val cpgVersion        = "1.6.10"
-val joernVersion      = "2.0.323"
+val joernVersion      = "2.0.324"
```
pandurangpatil, line 11: Once the comparison report looks positive try with the latest version `2.0.327`

pandurangpatil: @khemrajrathore Were you able to resolve the error?

pandurangpatil: Looks like the latest version of overflowdb is `1.192` did you try with `1.92` or `1.192`?

khemrajrathore: Yes, tried with `192` misspelt while copying

PR #1072, opened 2024-04-16
[PHP] Initial support for Collections

karan-batavia:
When using [Symfony](https://symfony.com/doc/current/routing.html), one can define routes and their corresponding controllers in `yaml` files. To tag sources inside these controllers, a separate pass is introduced which facilitates separation of concerns.


src/main/scala/ai/privado/languageEngine/php/tagger/collection/AnnotationsCollectionTagger.scala
```diff
@@ -0,0 +1,84 @@
...
+import ai.privado.model.{Constants, RuleInfo}
+import ai.privado.tagger.PrivadoParallelCpgPass
+import io.shiftleft.codepropertygraph.generated.Cpg
+import io.shiftleft.codepropertygraph.generated.nodes.{Annotation, Method}
+import io.shiftleft.semanticcpg.language.*
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable
+
+class AnnotationsCollectionTagger(cpg: Cpg, ruleCache: RuleCache) extends PrivadoParallelCpgPass[RuleInfo](cpg) {
```
khemrajrathore, line 14: The code present in this class, looks pretty much identical to other language's Annotation collection tagger. 

Create a base class with all the necessary functionality and override the differentiating piece in the language's tagger

src/main/scala/ai/privado/languageEngine/php/tagger/collection/ConfigCollectionTagger.scala
```diff
@@ -0,0 +1,188 @@
...
+            }
+        },
+        route.path
+      )
+    )
+
+    // The RuleInfo is intended to be empty
+    tagSources(
+      builder,
+      RuleInfo(
```
khemrajrathore, line 63: Why are we passing the empty ruleInfo here, please elaborate.

Also if we need to pass empty ruleInfo, better to create a helper function on top of `object RuleInfo` which facilitates this

src/test/scala/ai/privado/languageEngine/php/tagger/collection/AnnotationsCollectionTaggerTests.scala
```diff
@@ -0,0 +1,35 @@
...
+import ai.privado.model.{CatLevelOne, ConfigAndRules, Constants, SourceCodeModel}
+import ai.privado.rule.RuleInfoTestData
+import ai.privado.testfixtures.{PhpFrontendTestSuite, TestCpgWithPhp}
+import io.shiftleft.semanticcpg.language.*
+
+class AnnotationsCollectionTaggerTests extends PhpFrontendTestSuite {
+
+  val ruleCache =
+    RuleCache()
+  ruleCache.setRule(ConfigAndRules(sources = RuleInfoTestData.sourceRule, collections = collectionRules))
```
khemrajrathore, line 13:
```suggestion
  ruleCache.setRule(RuleInfoTestData.rule.copy(collections = collectionRules))
```

src/test/scala/ai/privado/languageEngine/php/tagger/collection/AnnotationsCollectionTaggerTests.scala
```diff
@@ -0,0 +1,35 @@
...
+          |      return $firstName;
+          |   }
+          | }
+          |>
+          |""".stripMargin)
+        .withRuleCache(ruleCache)
+
+      val List(editControllerMethod) = cpg.method("edit").l
+      editControllerMethod.tag.nameExact(Constants.catLevelOne).value.l shouldBe List(CatLevelOne.COLLECTIONS.name)
+      editControllerMethod.tag.name("COLLECTION_METHOD_ENDPOINT").value.l shouldBe List("/profile/edit")
```
khemrajrathore, line 32: Asserting where a given is a collection point is a repeated code, this can be modularized like below - https://github.com/Privado-Inc/privado-core/pull/1042/files#diff-ac08a96c9204201059395e6ea0a026e4dfd9b7fd4ac5462817f40723b2c22961
and can be used as `class AnnotationTagger extends PhpFrontendTestSuite with CollectionValidator`

src/test/scala/ai/privado/languageEngine/php/tagger/collection/AnnotationsCollectionTaggerTests.scala
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
```
khemrajrathore, line 34: Have seen it mutiple times that tags get added to collection point, but they are not reflected in privado.json. Please add test for that as well

src/main/scala/ai/privado/languageEngine/php/tagger/collection/ConfigCollectionTagger.scala
```diff
@@ -0,0 +1,188 @@
...
+    *   The alias of the controller.
+    * @return
+    *   The fully qualified controller class name if found, otherwise an empty string.
+    */
+  private def parseConfigToGetControllerFullName(fileName: String, alias: String): String = {
+    yaml.parser
+      .parse(better.files.File(fileName).contentAsString)
+      .map(parsedOp => {
+        parsedOp.findAllByKey(alias) match {
+          case List(someAlias) => someAlias.findAllByKey("class").map(_.asString.headOption.getOrElse(""))
```
khemrajrathore, line 175: Add comments to explain what exactly are we trying to read from the yaml file

src/test/scala/ai/privado/testfixtures/PhpFrontendTestSuite.scala
```diff
@@ -30,4 +31,26 @@ class TestCpgWithPhp(val fileSuffix: String, val language: Language.Value) exten
 }
 
 class PhpFrontendTestSuite(fileSuffix: String = ".php", language: Language.Value = Language.PHP)
-    extends PrivadoBaseTestFixture(() => new TestCpgWithPhp(fileSuffix, language)) {}
+    extends PrivadoBaseTestFixture(() => new TestCpgWithPhp(fileSuffix, language)) {
+
+  protected val collectionRules: List[RuleInfo] = List(
```
khemrajrathore, line 36: If the rule is used across, we need to define it in the `rule` package in test.
If it is used in a specific test cases, we can define it there itself.
If it is used in multiple test files, you can create a language package in rule, to store such repeated rules

khemrajrathore (commented):
- [x] Write comments for code blocks like Yaml file reading to explain, what exactly we anticipate to read
- [x] Extract common code of collection validation to a trait CollectionValidator, which can be used to assert
- [x] CollectionTagger's code is very much repeated and has scope of abstraction if we compare with other language CT, enforce no code duplication
- [x] Write test case on collections in privado.json

tuxology (dismissed):
LGTM 🌈

src/main/scala/ai/privado/languageEngine/php/processor/PhpProcessor.scala
```diff
@@ -89,7 +90,7 @@ class PhpProcessor(
       .withOutputPath(cpgOutput.toString)
       .withIgnoredFilesRegex(ruleCache.getExclusionRegex)
       .withPhpParserBin(PhpProcessor.parserBinPath)
-      .withDownloadDependencies(!config.skipDownloadDependencies)
+      .withDownloadDependencies(Try(!config.skipDownloadDependencies).getOrElse(false))
```
khemrajrathore, line 93: Why do we need a Try here?

khemrajrathore: For test cases where we want to download dependency or not we need to pass the respective flag in PrivadoInput, use `withPrivadoInput` while writing test case

tuxology: Tried, but reverted since we will have an issue where config is still not present in tests, but is expected still since it uses the same PhpProcessor now

khemrajrathore: @tuxology If it is needed in a test case, it needs to passed there, this helps in understanding the config needed to make the test case work. Ensure we pass PrivadoConfig with the flag set to all such test cases

khemrajrathore:
@tuxology Take a sync from Dev

khemrajrathore (approved):
LGTM

PR #1083, opened 2024-04-25
Apache Flink Support

khemrajrathore:
This PR adds support for Apache Flink
- support for Default sink connectors
- support for custom sink connectors

Pending Item- Running this for kotlin frontend and adding test case for the same

Privado PR - https://github.com/Privado-Inc/privado/pull/453


pandurangpatil (approved):
Looks good for now

PR #1084, opened 2024-04-28
Member filter in data element discovery

ankit-privado:
Filter out classes in the data element discovery of the audit report that have no member variables defined.


khemrajrathore (approved):
LGTM

PR #1088, opened 2024-05-02
Apache Airflow support

ankit-privado:
Added support for Apache Airflow and includes core and custom operator support. For core operators, we need to add rules but for custom operators, we find them by searching for classes that inherit from `BaseOperator`.


khemrajrathore (dismissed):
Lgtm

khemrajrathore (approved):
LGTM

PR #1090, opened 2024-05-02
Support for parsing hightouch repos

karan-batavia:
Includes, 
1. Parse YAML files from hightouch repositories.
2. Enhance SQL parser.
3. Refactor DefaultProcessor.
4. Add custom rules for hightouch which are generated/deserialized at runtime.
5. Some basic tests


src/main/resources/hightouch/rules/hightouch_rules.json
```diff
@@ -0,0 +1,2030 @@
+[
+  {
+    "id": "ThirdParties.SDK.1o",
```
khemrajrathore, line 3: Add Hightouch in rules, to not overlap with any actual rule passed from Privado rules
`ThirdParties.SDK.Hightouch.1o` like this

src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -537,15 +559,16 @@ object ScanProcessor extends CommandProcessor {
   private def processCpgWithDefaultProcessor(sourceRepoLocation: String, appCache: AppCache) = {
```
khemrajrathore, line 559: As we have removed the function call of this method from above, where else do we use this.

Instead of duplicating code, let's use this method above

src/main/scala/ai/privado/languageEngine/default/processor/DefaultProcessor.scala
```diff
@@ -23,165 +23,220 @@
...
+//        xtocpg,
+//        ruleCache,
+//        sourceRepoLocation,
+//        dataFlowCache,
+//        auditCache,
+//        s3DatabaseDetailsCache,
+//        appCache,
+//        propertyFilterCache = propertyFilterCache
+//      )
+//    msg
```
khemrajrathore, line 240: remove the commented code

pandurangpatil (approved):

PR #1096, opened 2024-05-03
Python perf op

pandurangpatil:


pandurangpatil:
Changes are largly with respect to the Python frontend so proceeding further with the merge.

PR #1101, opened 2024-05-06
Airflow UI blank issue fix

ankit-privado:
After creating the custom rule for operators, need to add rules into `RuleCache` so that when generating the `privado.json`  rule gets resolved.


src/main/scala/ai/privado/languageEngine/python/tagger/sink/AirflowOperatorSinkPass.scala
```diff
@@ -33,24 +33,32 @@ class AirflowOperatorSinkPass(cpg: Cpg, ruleCache: RuleCache) extends PrivadoPar
...
+        false,
+        "",
+        Map(),
+        NodeType.REGULAR,
+        "",
+        CatLevelOne.SINKS,
+        Constants.third_parties,
+        Language.PYTHON,
+        Array()
+      )
```
khemrajrathore, line 61: There is also a field `isGenerated` set that to `true`

ankit-privado: done.

khemrajrathore (approved):
Lgtm

PR #1102, opened 2024-05-06
fix failing flink test case

khemrajrathore:
- Introduces `dumpCpgAtBasePath`, which if `true` will write the cpg at the sourceRepoPath being scanned and doesn't create a `.privado` folder
- Fixes the kotlin test case failure when using `KotlinFrontendTestSuite`, The test case are failing as we create a `.privado` folder before parsing the code, which promptly the current subdirectory fetching logic to only consider `.privado` folder content for kotlin scans which inturn have 0 kotlin file, resulting into empty CPG


pandurangpatil:
@khemrajrathore I recommend fixing this in the Language engine rather than here. Let's try that option, if it takes time, we have the solution.

khemrajrathore:
We decided on a different apporach, closing this PR

PR #1108, opened 2024-05-08
Support for forcing scan with the langauge

pandurangpatil:


src/main/scala/ai/privado/entrypoint/ScanProcessor.scala
```diff
@@ -412,7 +411,7 @@ object ScanProcessor extends CommandProcessor {
                     appCache,
                     propertyFilterCache = propertyFilterCache
                   ).processCpg()
-              case language if language == Languages.JSSRC =>
+              case language if language == Languages.JSSRC || language == Languages.JAVASCRIPT =>
```
khemrajrathore, line 414: We can get rid of the Joern's `Languages` in our core logic, and map the language to our Language in the above `guessLanguage` result. That should avoid the confusion of whether to Use `Languages` or Privado's Language

pandurangpatil: done

khemrajrathore (approved):
LGTM

PR #1124, opened 2024-05-15
Added thread dump after regular interval of time

pandurangpatil:
Added thread dump after regular interval of time default to 10 mins and if the avg cpu utilisation is lower than the 50% for given stage.

### DIFF
--- a/src/main/scala/ai/privado/entrypoint/CommandParser.scala
+++ b/src/main/scala/ai/privado/entrypoint/CommandParser.scala
@@ -23,8 +23,8 @@
 package ai.privado.entrypoint
 
 import ai.privado.metric.MetricHandler
-import ai.privado.utility.StatsRecorder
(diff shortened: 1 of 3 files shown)


khemrajrathore (changes requested):
I see couple of integer values used across, Let's create constant for these and use, this will help in having the values configured at a single place and will reduce the time and bugs if in future we want to change them

Also add a comment what this integer values point to like when we set x = 60000, add a comment time in seconds etc

pandurangpatil:
> I see couple of integer values used across, Let's create constant for these and use, this will help in having the values configured at a single place and will reduce the time and bugs if in future we want to change them
> 
> Also add a comment what this integer values point to like when we set x = 60000, add a comment time in seconds etc

done

khemrajrathore (approved):
LGTM

PR #1127, opened 2024-05-18
fix: APIs should not be tagged as collection

tuxology:
* use updated joern which populates fullname in annotation
 * skip retrofit annotations when tagging collections

### DIFF
--- a/src/test/scala/ai/privado/languageEngine/kotlin/tagger/sink/framework/retrofit/RetrofitAPITaggerTest.scala
+++ b/src/test/scala/ai/privado/languageEngine/kotlin/tagger/sink/framework/retrofit/RetrofitAPITaggerTest.scala
@@ -1,0 +1,83 @@
+package ai.privado.languageEngine.kotlin.tagger.sink.framework.retrofit
+
+import ai.privado.testfixtures.KotlinFrontendTestSuite
+import io.shiftleft.semanticcpg.language.*
+import ai.privado.cache.RuleCache
+import ai.privado.model.*
+import ai.privado.rule.RuleInfoTestData
+import ai.privado.tagger.sink.api.APIValidator
+import ai.privado.entrypoint.PrivadoInput
+import ai.privado.model.{CatLevelOne, Constants, SourceCodeModel}
+
+class RetrofitAPITaggerTest extends KotlinFrontendTestSuite {
+
+  val collectionRule = List(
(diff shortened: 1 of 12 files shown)


src/test/scala/ai/privado/languageEngine/kotlin/tagger/sink/framework/retrofit/RetrofitAPITaggerTest.scala
```diff
@@ -0,0 +1,83 @@
...
+import ai.privado.cache.RuleCache
+import ai.privado.model.*
+import ai.privado.rule.RuleInfoTestData
+import ai.privado.tagger.sink.api.APIValidator
+import ai.privado.entrypoint.PrivadoInput
+import ai.privado.model.{CatLevelOne, Constants, SourceCodeModel}
+
+class RetrofitAPITaggerTest extends KotlinFrontendTestSuite {
+
+  val collectionRule = List(
```
karan-batavia, line 14: Is there a base class where these rules could be added? If these are not specific to this test, can we move them there?

karan-batavia (changes requested):

src/test/scala/ai/privado/rule/CollectionRuleTestData.scala
```diff
@@ -0,0 +1,27 @@
+package ai.privado.rule
+
+import ai.privado.model.{CatLevelOne, Constants, FilterProperty, Language, NodeType, RuleInfo}
+
+object CollectionRuleTestData {
```
karan-batavia, line 5: Shouldn't this be under the package `ai.privado.rule.kotlin`? Small change but will ensure better consistency, as this rule is only for kotlin.

karan-batavia (changes requested):

karan-batavia:
One small comment, rest looks good. Also, I see missing flows for python? Can we verify those before merging?

src/test/scala/ai/privado/rule/CollectionRuleTestData.scala
tuxology: Ideally this should work for any language. I'll redefine the rule to make it work for any language so we don't need to move it to any any other package

karan-batavia (approved):
LGTM

khemrajrathore (approved):
LGTM

PR #1143, opened 2024-05-29
Ruby parser timeout

pandurangpatil:
We made changes to add timeout while parsing the file. In earlier case when David had tried the parser timeout. Every file was getting parsed inside AstCreationPass thread. In order to add the timeout, he tried spawning another thread (Future). I suspect the inconsistency issue that we witnessed at that time, might have been related to it. 

With the updated structure recently, it seems we have segregated the parsing from the AstCreation. Where all the files are parsed first in the separate flow and then parsed files are passed to AstCreationPass. With this we added the timeout mechanism for each parsing task. 

I have executed three scans for the same repo having files which run longer in parsing stage. So far results have been consistent. 
I will be testing more for consistency and share the update on this PR


khemrajrathore (approved):
LGTM

pandurangpatil:
Verified the consistency in results across multiple runs for repository retailer-platform-api and the results were consistent. So merging this PR.

PR #1146, opened 2024-05-30
Release PR

pandurangpatil:
* Made changes to add timeout for parser

* error log

* Change to identify parser timedout files

* Error handling

* logged error and timeout file count separately

* log change

* timeout change

Earlier timeout was configured for execution of all the tasks. Made changes to have timeout only for individual tasks

* proper timeout handling with future

* made changes to use predfined executor service pool for future


ankit-privado (dismissed):
LGTM

PR #1148, opened 2024-06-03
Change `updateDependencies.sh` to resolve versions from GH packages

karan-batavia:
Changes: 
1. Use `jq` instead of `xmllint` to parse response. ( GH Packages works with json )

PS: Would require a GH token in the environment to run this because the package registry is private to an organization.

### DIFF
--- a/updateDependencies.sh
+++ b/updateDependencies.sh
@@ -12,30 +12,38 @@ check_installed() {
 check_installed curl
 
 # check if xmllint is installed
-if type xmllint > /dev/null; then
-  USE_XMLLINT=1 #true
+if type jq > /dev/null; then
+  USE_JQ=1 #true
 else
-  echo "warning: xmllint is not installed - will try with 'grep' as a fallback..."
-  USE_XMLLINT=0 #false
+  echo "warning: jq is not installed - will try with 'grep' as a fallback..."
+  USE_JQ=0 #false
 fi
 
 declare -A repos=(
-  [cpg]=https://repo1.maven.org/maven2/io/shiftleft/codepropertygraph-schema_3
-  [joern]=https://repo1.maven.org/maven2/io/joern/console_3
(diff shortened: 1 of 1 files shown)


updateDependencies.sh
```diff
@@ -11,31 +11,39 @@ check_installed() {
...
 
-  if (( $USE_XMLLINT ))
+  if (( $USE_JQ ))
   then
-    curl $CURL_PARAMS | xmllint --xpath "/metadata/versioning/latest/text()" -
+    curl $CURL_PARAMS -L \
+                       -H "Accept: application/vnd.github+json" \
+                       -H "Authorization: Bearer ${TOKEN}" \
+                       -H "X-GitHub-Api-Version: 2022-11-28" \
+                       "$REPO_URL" | jq '[.[]|select(.name)][0]|.name' | sed 's/[",[:space:]]//g' # get the latest version, and extract the version number
```
khemrajrathore, line 40: Do some testing here, to make sure we are covering scenarios while greping

updateDependencies.sh
```diff
@@ -11,31 +11,39 @@ check_installed() {
...
   else
-    curl $CURL_PARAMS | grep '<latest>' | sed 's/[ ]*<latest>\([0-9.]*\)<\/latest>/\1/'
+    curl $CURL_PARAMS -L \
+                   -H "Accept: application/vnd.github+json" \
+                   -H "Authorization: Bearer ${TOKEN}" \
+                   -H "X-GitHub-Api-Version: 2022-11-28" \
+                   "$REPO_URL" | grep -m 1 "\"name\":" | sed 's/[",[:space:]]//g' | awk '{split($0,a,":"); print a[2]}' # use grep to get the first occurrence of the name key which contains the latest version
```
khemrajrathore, line 46: Do some testing here, to make sure we are covering scenarios while greping

khemrajrathore (approved):
LGTM

PR #1166, opened 2024-06-10
Slack Env comparison fix

ankit-privado:
Added Slack Token to workflow's ubuntu instance.

### DIFF
--- a/.github/workflows/comparison-results.yml
+++ b/.github/workflows/comparison-results.yml
@@ -58,7 +58,7 @@ jobs:
 
       # langauge specific repository file
       - name: Run the script for ${{github.head_ref}} and ${{github.base_ref}}
-        run: export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} && export SLACK_TOKEN=${{ secrets.SLACK_TOKEN }} && export SLACK_CHANNEL_1=${{ secrets.SLACK_CHANNEL_1 }} && cd ./temp/standalone-monitoring-stability && pip install -r requirements.txt && python3 ./run.py -r ./repos/${{matrix.language}}.txt -b ${{github.base_ref}} -h ${{github.head_ref}} -bcr ${{ github.event.pull_request.base.repo.html_url }} -hcr ${{ github.event.pull_request.head.repo.html_url }} -guf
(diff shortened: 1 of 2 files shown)


.github/workflows/comparison-results.yml
```diff
@@ -58,7 +58,7 @@ jobs:
 
       # langauge specific repository file
       - name: Run the script for ${{github.head_ref}} and ${{github.base_ref}}
-        run: export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} && cd ./temp/standalone-monitoring-stability && pip install -r requirements.txt && python3 ./run.py -r ./repos/${{matrix.language}}.txt -b ${{github.base_ref}} -h ${{github.head_ref}} -bcr ${{ github.event.pull_request.base.repo.html_url }} -hcr ${{ github.event.pull_request.head.repo.html_url }} -guf
+        run: export GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }} && export SLACK_TOKEN=${{ secrets.SLACK_TOKEN }} && cd ./temp/standalone-monitoring-stability && pip install -r requirements.txt && python3 ./run.py -r ./repos/${{matrix.language}}.txt -b ${{github.base_ref}} -h ${{github.head_ref}} -bcr ${{ github.event.pull_request.base.repo.html_url }} -hcr ${{ github.event.pull_request.head.repo.html_url }} -guf
```
karan-batavia, line 61: Can we use the `env` key for this? It is much cleaner, and will help if we want to add more environment variables.

karan-batavia (changes requested):

.github/workflows/comparison-results.yml
ankit-privado: Yes right, need to change at multiple places like where its fetched on comparison repo.

karan-batavia: Yes, discovered this while working on the pipeline. Should be a small change.

HiteshMah-Jan (approved):

ankit-privado:
As discussed with @karan-batavia , the fix for the issue mentioned above is not required now since we have already switched to using Privado's own Joern instead of the previously publicly available Joern. Therefore, im closing this PR for now.

PR #1171, opened 2024-06-12
Increment lineNumber if file is .cs

karan-batavia:
These will be reverted once the actual root cause is fixed. 
Ticket for the fix: https://privado-ai.atlassian.net/jira/software/c/projects/LE/boards/8?assignee=637b03a85fc160544e1898c4&selectedIssue=LE-1384


tuxology (approved):

ankit-privado (approved):
LGTM

PR #1183, opened 2024-06-19
Set `suppressSubstageFlag` to true except in the `java` frontend

karan-batavia:


khemrajrathore (commented):
What does this flag do?

karan-batavia:
> What does this flag do?

@khemrajrathore This flag is supposed to suppress logs after a certain point if set to `true`. In this case, the flag is by default to `true`, meaning it will start logging after it is set to `false` which in turn means that the stats will be logged for custom privado passes only.

PR #1211, opened 2024-07-05
linking any mathod which takes property key as parameter

pandurangpatil:


src/main/scala/ai/privado/languageEngine/java/passes/config/JavaPropertyLinkerPass.scala
```diff
@@ -144,7 +144,6 @@ class JavaPropertyLinkerPass(cpg: Cpg) extends PrivadoParallelCpgPass[JavaProper
 
   private def matchingLiteralsInGetPropertyCalls(propertyName: String): List[Literal] = cpg.literal
     .codeExact("\"" + propertyName + "\"")
-    .filter(_.inCall.name(".*getProperty").nonEmpty)
```
khemrajrathore, line 147: This directly matches on the literal. Reading the PR title `linking any mathod which takes property key as parameter` There is no check for literal being a argument to a call or parameter to a method

khemrajrathore (approved):
LGTM

PR #1213, opened 2024-07-06
Revert "Use alpine latest and tighten the threshold of image scanner"

dbMundada:
Reverts Privado-Inc/privado-core#1210


pandurangpatil:
@nikvin15 @khemrajrathore reverting these changes https://github.com/Privado-Inc/privado-core/pull/1210 as we were not sure about the final results of the testing of these changes.

PR #1217, opened 2024-07-08
refactor and make sure to use the trait as base for all ApiSinkTagger

khemrajrathore:
The PR adds
- support of adding inference API rules across languages
- support for taking API sink via methodFullName regex
- Refactor to make this available across


ankit-privado (approved):
LGTM

PR #1222, opened 2024-07-09
filter out already tagged api sinks

khemrajrathore:
Context - We run inference tagger to tag API, and after than we run a API Tagger.
API's already tagged by inference, shouldn't process with API Tagger, so we have added a filtering in `sinkTagger` utility which was currently only applied for Java. But as we have opened up the inference tagger for other language as well, having it at a common place is ideal.


pandurangpatil (approved):
Looks good to me

PR #1243, opened 2024-08-07
fix - file path stored in file node should be relative

khemrajrathore:


src/main/scala/ai/privado/passes/PropertyParserPass.scala
```diff
@@ -354,7 +355,8 @@ class PropertyParserPass(
   }
 
   private def addFileNode(name: String, builder: BatchedUpdate.DiffGraphBuilder): NewFile = {
-    val fileNode = NewFile().name(name)
+    val relativeFileName = Path.of(projectRoot).relativize(Path.of(name)).toString
+    val fileNode         = NewFile().name(relativeFileName)
```
karan-batavia, line 359: Is this in sync with what joern has? We wouldn't want to be out of sync with them where only the property files have relative paths, and others don't.

khemrajrathore: Yes, we have relative paths across

karan-batavia (approved):

src/main/scala/ai/privado/passes/PropertyParserPass.scala
karan-batavia: Approved. You might also want to check if we missed any custom nodes for relative paths.

PR #1250, opened 2024-08-16
Dynamic rule merging

ankit-privado:
Dynamic rule create by 3ps discovery will be merged with existing rule with defined logic.

### DIFF
--- a/src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
+++ b/src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
@@ -3,9 +3,11 @@ package ai.privado.entrypoint
 import ai.privado.model.{ConfigAndRules, FilterProperty, RuleInfo}
 import org.slf4j.LoggerFactory
 
+import scala.collection.mutable
+import scala.collection.mutable.Map
 import scala.collection.mutable.ListBuffer
 
-object DynamicRuleMerger {
+trait DynamicRuleMerger {
 
   private val logger = LoggerFactory.getLogger(this.getClass)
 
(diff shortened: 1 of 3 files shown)


src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
```diff
@@ -0,0 +1,44 @@
...
+import ai.privado.model.{ConfigAndRules, RuleInfo}
+import org.slf4j.LoggerFactory
+
+import scala.collection.mutable.ListBuffer
+
+object DynamicRuleMerger {
+
+  private val logger = LoggerFactory.getLogger(this.getClass)
+
+  def mergeDynamicRuleSinkForDependencyDiscovery(
```
pandurangpatil, line 12: Any reason to create this as separate Object `DynamicRuleMerger` instead of adding this method `mergeDynamicRuleSinkForDependencyDiscovery` inside `RuleProcessor`?

pandurangpatil: In my opinion, it should be part of `RuleProcessor` class.

ankit-privado: Yes we can. My thought was to separate the merging utility from the `ruleProcessor` because, in the future, there will be more merging logic for `storage`, `API`, and their special cases.

src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
```diff
@@ -0,0 +1,44 @@
...
+          (rule.domains.contains(externalDomain) || rule.name == externalRuleName) && rule.id.contains(
+            "ThirdParties.SDK"
+          ) && rule.filterProperty != FilterProperty.CODE
+        } match {
+          case index if index >= 0 =>
+            val matchingRuleInfo = updatedRules(index)
+            val updatedRule      = matchingRuleInfo.copy(patterns = matchingRuleInfo.patterns ++ externalRule.patterns)
+            updatedRules.update(index, updatedRule)
+          case _ =>
+            updatedRules += externalRule
```
khemrajrathore, line 33: @ankit-privado I have a question regarding performance here, for every external rules, we will be iterating over all internal rules, better to use `Map` instead of `ListBuffer` and the key can be the rule name
cc - @pandurangpatil

src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
khemrajrathore: I propose to create a `trait DynamicRuleMerger` and we can use the functions of this class as `trait RuleProcessor extends DynamicRuleMerger`

ankit-privado: Yes right using `trait` would fulfil out requirement.

ankit-privado: done.

src/main/scala/ai/privado/entrypoint/DynamicRuleMerger.scala
ankit-privado: yes done.

khemrajrathore (approved):
Lgtm

PR #1265, opened 2024-08-27
Dynamic Rule Merger Fix

ankit-privado:
When we merge an external rule with an internal one, it's possible that both merging rules don't have the same ID. In cases where the domain or name matches, we merge the external rule with the internal one, but when tagging the dependency node, we only have the external rule ID. So, it's important to map which external rule was merged with which internal rule.

We have created the `DynamicMergerRuleMap` to store the merging information.


pandurangpatil (approved):
Looks good to me.

PR #1268, opened 2024-09-05
Dev

karan-batavia:


.github/workflows/comparison-results.yml
```diff
@@ -6,22 +6,35 @@ on:
     branches:
       - "**"
 
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  CORE_AT: ${{ secrets.CORE_AT }}
```
khemrajrathore, line 11: what is the use of this env variable `CORE_AT`?

